PR #94, opened 2024-06-26
feat(utils): support new naming convention

andypf:
do not add @juno prefix for packages which start with @ in their name


ArtieReus (approved):
lgtm

PR #109, opened 2024-07-09
chore(renaming): rename libs folder in packages

andypf:


edda (approved):

apps/template/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "template",
+  "name": "@cloudoperators/juno-app-template",
```
ArtieReus, line 2: can we also update the package-lock file

packages/template-lib/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "template-lib",
+  "name": "@cloudoperators/juno-template-lib",
```
ArtieReus, line 2: please update also package-lock.json

packages/communicator/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "communicator",
+  "name": "@cloudoperators/juno-communicator",
```
ArtieReus, line 2: please update also package-lock.json

ArtieReus (changes requested):

ArtieReus (approved):
lgtm

PR #110, opened 2024-07-10
feat(url-state-provider): URI-Safe, and readable encoding and compression of various data types and nested objects

TilmanHaupt:
The new custom encoding and decoding with following benifits:

- URI-safe encoding with RFC 3986 compatible characters 
- Support for various data types (like numbers, regex, strings, array ())
- The encoding can handle nested objects and arrays, maintaining their structure through the encoding and decoding process
- human readable
- Encoding supports also lz-compression and base64, while maintaining structure and data types
- Special characters in strings can not crash the decoding anymore like with JURI  are encodes with ~ or % 
- spaces are now encoded with a +  
- Error handling: The code includes error checking and provides options for null returns on decoding errors
- 100% test coverage for the encoding and decoding
-  fully implement all features of the encoding in url-state-provider
 - update READ.ME

### DIFF
--- a/packages/url-state-provider/jest.config.js
+++ b/packages/url-state-provider/jest.config.js
@@ -4,6 +4,7 @@
  */
 
 module.exports = {
+  testEnvironment: "jsdom",
   transform: { "\\.[jt]sx?$": "babel-jest" },
   transformIgnorePatterns: [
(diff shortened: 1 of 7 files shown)


andypf (approved):
lgtm

PR #111, opened 2024-07-11
chore(juno): add ui-components and turbo repo first draft 

barsukov:
This is the first monorepo attmeption and this is the task that could run

- turbo build
- turbo dev

And we should update Readme and check out ui-components and npm. In the end we will publish them separate under our registry.


andypf (approved):
lgtm

PR #113, opened 2024-07-15
Andypf communicator preparation

andypf:


apps/example/esbuild.config.js
```diff
@@ -198,7 +198,7 @@ const build = async () => {
 
       let { host, port } = await ctx.serve({
         host: "0.0.0.0",
-        port: parseInt(process.env.APP_PORT || process.env.PORT || 3000),
+        port: parseInt(process.env.APP_PORT || process.env.PORT || 3020),
```
barsukov, line 201: I will keep the example app on 3000 tbh cause it is a main default entry point I would say and more clear for everyone, and change communicator port

PR #117, opened 2024-07-15
feat(docs): add a decision document for monorepo

andypf:


barsukov (approved):

edda (changes requested):
Can you add a sentence about how the `*` internal dependency will be resolved when publishing to npm?

docs/decision_monorepo_tooling.md
```diff
@@ -0,0 +1,30 @@
...
+1. **Dependency Management:**
+
+   - Internal dependencies within the monorepo will be marked with `*` to streamline development processes.
+
+2. **Tool Adoption:**
+
+   - Adoption of `turborepo` is proposed to optimize task management and leverage caching for performance improvements.
+
+3. **Release Management:**
+   - The `changeset` CLI tool will be used to manage versioning and changelogs for packages within the monorepo. It will streamline the process of tracking changes, updating versions, and generating changelogs for releases.
```
barsukov, line 30: I think it would be nice to make it consistent with other points so same new lines in between like in point 1 and 2

edda (approved):

Issue #119, opened 2024-07-15
[EPIC] (setup) Integrate typescript to juno core


barsukov:
# Epic: Transition Apps and Packages to TypeScript

## Description

Transition all apps and packages in the `packages` folder from JavaScript to TypeScript to enhance code quality and maintainability. This epic will focus on ensuring a smooth migration with minimal disruption.

## Goals
- Setup Typescript package with a proper typecheck on turbo task(Evaluate tsup runner)
- Convert all JavaScript files in the `packages` folder to TypeScript
- Ensure compatibility and functionality of each package post-transition
- Update build, test, and CI/CD pipelines to support TypeScript

## Benefits

- Improved type safety and code clarity
- Better developer experience with enhanced tooling
- Early error detection through static type checking

## Tasks

- [x] Audit `packages` folder for TypeScript compatibility
- [x] Configure TypeScript for Monorepo global
- [ ] Configure TypeScript for each package
- [ ] Convert each package to TypeScript
  - [x] #168
  - [x] communicator
  - [x] #196
  - [x] #197
  - [x] #198
  - [x] #199
  - [x] #163
  - [x] #201
  - [x] #202
- [ ] Update unit tests for each package to TypeScript
- [ ] Update documentation to reflect TypeScript usage for each package

## Acceptance Criteria

- All packages in the `packages` folder are successfully converted to TypeScript
- Build and deployment pipelines compile and deploy TypeScript code without issues
- Updated documentation and guidelines for contributing TypeScript code are available
- All tests pass with TypeScript integration for each package

edda:
All packages migrated to Typescript

PR #162, opened 2024-07-31
chore(ui): UI components typescript integration

barsukov:
# Summary
This is a first integration of typescript in a mixing approach that we have now  supported javascript system.
# Changes Made

- Integrate typescript config to ui-components through rollup
- integrate typescript config to storybook
- Integrate ts lint for Typescript related files
- Integrate vitest and typescript support in testing.

# Related Issues
- #119 

# Testing Instructions

1. `npm i`
2. `turbo test --filter=@cloudoperators/juno-ui-components`
3. `turbo build --filter=@cloudoperators/juno-ui-components`

# Checklist


apps/example/tailwind.config.js
```diff
@@ -5,7 +5,7 @@
 
 module.exports = {
   presets: [
-    require("@cloudoperators/juno-ui-components/build/lib/tailwind.config"), // important, do not change
+    require("@cloudoperators/juno-ui-components/dist/lib/tailwind.config"), // important, do not change
```
edda, line 8: Renaming the build dir is a breaking change for all apps that depend on `juno-ui-components`. In this repo it's only the example and template app but we have all kinds of apps out in the world that will break if we make this change because they won't have access to the tailwind config that we provide to apps. This means they won't have access to our colors and other custom classes anymore.

packages/ui-components/src/components/AppShell/AppShell.test.tsx
```diff
@@ -0,0 +1,175 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+// // TODO: When we the other components migrate to TS we will migrate this test in the end and delete this dummy
```
edda, line 5: Is it really necessary to comment out all the tests for this component? We have an issue planned for this sprint which will touch the AppShell. It's not great if all the tests are turned off.

Maybe it would have been easier to first migrate a component that's not used in the tests of a component that hasn't been migrated yet?

packages/ui-components/src/components/PageHeader/PageHeader.component.tsx
```diff
@@ -4,8 +4,7 @@
  */
 
 import React from "react"
-import PropTypes from "prop-types"
-import SAPLogo from "../../img/sap_logo.svg"
+import { SapSvg } from "../../svg/SapSvg"
```
edda, line 7: Why the name change? The old name told me exactly what it is, the new name tells me that it's an SVG but I lose the semantic information.

packages/ui-components/src/svg/SapSvg.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react"
+
+interface SapSvgProps {
+  className?: string
+  alt?: string
+}
+
+export const SapSvg: React.FC<SapSvgProps> = ({ className, alt }) => {
```
edda, line 8: Is this the only way to use SVGs as components in TypeScript? This means we always have to take the extra step of turning every image into a TypeScript component. In addition you can't view them in an image editor or other context where you want to see what the image looks like. I would like to be able to keep the image source files as proper images. If TypeScript requires images to be TypeScript components then can we write some sort of wrapper that takes an image and turns it into a component?

edda (changes requested):

apps/example/tailwind.config.js
barsukov: Gotcha, good catch! I thought about it too, I think it does not really matter build or dist, just on all examples in ts world I found dists, I thought if we will switch to typescript to make everything consistent. But I will make then build folder everywhere.

packages/ui-components/src/components/AppShell/AppShell.test.tsx
barsukov: Make sense. Could you suggest one that is easiest and not depends to others(not imported), that I can also easily check on a Example or Storybook page?

packages/ui-components/src/components/PageHeader/PageHeader.component.tsx
barsukov: Good point. I will change it.

packages/ui-components/src/svg/SapSvg.tsx
barsukov: I have an idea(how I already solved svgs problems before), we could discuss it on a 1:1 call.

edda (approved):

PR #178, opened 2024-08-06
feat(ui) Appshell: render TopNavigation in embedded node if passed

### LINKED ISSUE #99: feature(ui): AppShell should not hide TopNavigation in embedded mode
If you propogate `topNavigation` in the AppShell component like: `<AppShell topNavigation={navigation} embedded={bool}>` the Navigation is not shown when there is also the embedded prop set to true.

franzheidl:
# Feature: render TopNavigation as passed in embedded mode

Previously, it was impossible to render a `TopNavigation` in the `AppShell` in `embedded` mode. With this change a `TopNavigation` component can be passed as a prop in the `navigation` slot. Fixes #99 .

# Related Issues

- Issue 1: #99 

# Testing Instructions

`npm run test AppShell`


edda (approved):

packages/ui-components/src/components/AppShell/AppShell.test.js
```diff
@@ -107,6 +107,19 @@ describe("AppShell", () => {
     expect(screen.queryByText("My Page Footer")).not.toBeInTheDocument()
   })
 
+  test("renders a TopNavigation as passed in embedded mode", async () => {
```
barsukov, line 110: Just want to point out async rule that I found it out difficult on a future ts migration.

edda: Good point. Unfortunately I think it's everywhere due to copy/paste. But at least we can remove the unecessary ones now every time we touch the tests.

PR #192, opened 2024-08-08
chore(ui): wrapper class name new

franzheidl:
# Add `wrapperClassName` prop

Add a prop to allow for adding a custom `wrapperClassName` to the wrapping element of some components. 

# Changes Made

`wrapperClassName` prop added to the following components:
- ComboBox
- DateTimePicker
- NativeSelect
- Select
- Switch
- Textarea
- TextInput


packages/ui-components/src/components/ComboBox/ComboBox.test.js
```diff
@@ -533,6 +533,12 @@ describe("ComboBox", () => {
     expect(cbox).toHaveValue("option 2")
   })
 
+  test("renders a wrapperClassName to the outer wrapping <div> element", async () => {
```
barsukov, line 536: Do you need to make this test async? cause the lint going to fail with error async no await(when we move this class to ts)

packages/ui-components/src/components/NativeSelect/NativeSelect.test.js
```diff
@@ -112,6 +112,12 @@ describe("NativeSelect", () => {
     expect(onClickSpy).not.toHaveBeenCalled()
   })
 
+  test("renders a wrapperClassName to the outer wrapping element", async () => {
```
barsukov, line 115: Same here.

packages/ui-components/src/components/Textarea/Textarea.test.js
```diff
@@ -131,6 +131,12 @@ describe("Textarea", () => {
     expect(onChangeSpy).not.toHaveBeenCalled()
   })
 
+  test("renders a wrapperClassName to the wrapping <div> element", async () => {
```
barsukov, line 134: Same here.

packages/ui-components/src/components/TextInput/TextInput.test.js
```diff
@@ -166,6 +166,12 @@ describe("TextInput", () => {
     expect(onChangeSpy).not.toHaveBeenCalled()
   })
 
+  test("renders a wrapperClassName to the outer wrapper element", async () => {
```
barsukov, line 169: Same here

packages/ui-components/src/components/Switch/Switch.test.js
```diff
@@ -127,6 +127,12 @@ describe("Switch", () => {
     expect(screen.getByRole("switch")).toHaveClass("juno-switch-invalid")
   })
 
+  test("renders a wrapperClassName to the outer wrapper element", async () => {
```
barsukov, line 130: Same here

packages/ui-components/src/components/Select/Select.test.js
```diff
@@ -581,3 +581,13 @@ describe("Select", () => {
     expect(toggle).toHaveTextContent("Option 1")
   })
 })
+
+test("renders a wrapperClassName to the outer wrapping element", async () => {
```
barsukov, line 585: Same here.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.test.js
```diff
@@ -941,6 +941,12 @@ describe("DateTimePicker", () => {
     expect(mockOnMonthChange).toHaveBeenCalled()
   })
 
+  test("renders a wrapperClassName to the outer wrapping element", async () => {
```
barsukov, line 944: Same here.

barsukov (changes requested):
Great job, just one little request that all tests that use async should be also using the await(this rule is on if we use a tslint) the problem that we will need to fix it later, so to avoid it just remove async or add await :)

barsukov (approved):
LGTM

Issue #160, opened 2024-07-30
[TASK]:  @cloudoperators/juno-ui-components storybook upgrade from 7.x.x to 8.x.x


barsukov:
**Task Description**
We need to do an ugrade of storybook version, check first which preparation need to be done

**Sub-tasks**

- [x] Do upgrade in packages json
- [x]  See breaking changes
- [ ] Solve if it is a breaking change, do upgrade
- [x] Our self-developed plugin for theme switching doesn't work out of the box with v8.x and needs to be fixed

**Additional Context**
Please have a look for the info here:
[Storybook migration guide
](https://storybook.js.org/docs/migration-guide)

PR #208, opened 2024-08-14
chore(ui): update storybook version to 8.2.7

gjaskiewicz:
Update storybook version to 8.2.7

## Related issue
[#160](https://github.com/cloudoperators/juno/issues/160)


packages/ui-components/package.json
```diff
@@ -37,17 +37,18 @@
     "@rollup/plugin-commonjs": "^24.0.0",
     "@rollup/plugin-node-resolve": "^15.1.3",
     "@rollup/plugin-typescript": "11.1.6",
-    "@storybook/addon-docs": "^7.6.20",
```
edda, line 40: It looks really good to me except for one thing:
The WiP/DateTimePicker story is broken because of an error with flatpickr (the headless date time picker package that we use internally). Error message is:
```
flatpickrInstanceRef.current.destroy is not a function
```

edda (changes requested):

packages/ui-components/package.json
gjaskiewicz: I think I have idea what is wrong. I'll try to resolve it today

gjaskiewicz: I fixed it and added commit to this PR

edda (approved):
Looking good +1

PR #287, opened 2024-09-02
fix(build): fix package output directory issues

andypf:
# Summary

The recent changes to the main and module entries in the package.json files for communicator, k8sClient, and oauth broke these packages. Since the build is done with Vite and the default output folder is dist, there was a discrepancy between the settings and the actual output folders.

This was fixed by adding the --outDir option to the Vite CLI.

# Changes Made

- Add `--outDir` option to vite CLI in
  - oauth
  - communicator
  - k8sclient

# Checklist

### DIFF
--- a/packages/k8s-client/package.json
+++ b/packages/k8s-client/package.json
@@ -16,7 +16,7 @@
     "lint": "eslint",
     "typecheck": "tsc --noEmit",
     "dev": "vite",
-    "build": "vite build",
+    "build": "vite build --outDir build",
     "serve": "vite preview",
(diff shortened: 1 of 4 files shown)


packages/k8s-client/package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "eslint",
     "typecheck": "tsc --noEmit",
     "dev": "vite",
-    "build": "vite build",
+    "build": "vite build --outDir build",
```
barsukov, line 19: It is already set in vite conifg why do we need it?

barsukov (changes requested):
lets have a call on this PR, I do not think we need this changes.

PR #290, opened 2024-09-02
fix(ci): restore accidentally deleted shared workflow

ArtieReus:
# Summary

This PR restores a workflow that was accidentally modified. As the workflow is currently inactive, restoring it ensures that both the Greenhouse and Greenhouse-Extensions projects can continue to check the Juno applications.

### DIFF
--- a/.github/workflows/pipeline-npm-package.yaml
+++ b/.github/workflows/pipeline-npm-package.yaml
@@ -1,31 +1,86 @@
-name: Juno NPM Package Pipeline
+# Run it locally with act (https://github.com/nektos/act)
+#  1. Install act:
+#     `brew install act`
+#  2. Create a .secret file with the following content:
+#     `GITHUB_TOKEN=your_github_token`
+#  PULL REQUEST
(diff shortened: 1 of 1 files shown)


.github/workflows/pipeline-npm-package.yaml
```diff
@@ -1,31 +1,86 @@
...
+#     `{"inputs": {"paths": "packages/juno-ui-components apps/exampleapp", "node": "20.x", "runs-on": "ubuntu-latest"}}`
+#  2. Run the following command:
+#     `act workflow_call -e act_workflow_call_ci.json -W .github/workflows/check-changes-npm-package.yaml`
 
-on:
-  pull_request:
-    types: [opened, synchronize] # Triggers when a PR is opened or updated
+name: Detect Changes on the given Paths
 
-env:
```
barsukov, line 7: I would still keep this one, since it does not really have any side effects on other projects

barsukov (changes requested):
One little suggestion

barsukov (approved):

PR #301, opened 2024-09-04
fix(ui): forward button title to icon

VoigtS:
# Summary

The previous implementation always inserted the icon name into its title, disregarding the button title.

# Changes Made

- Forward the button title to the Icon if one should be displayed
- Bump the version.

image

In this screenshot, the Icon was hovered with a custom title set on the button.

# Testing Instructions
Open the storybook. Add a custom title on the button, hover over the Icon for a few seconds until the title shows up.


packages/ui-components/src/components/Button/Button.component.js
```diff
@@ -130,6 +130,7 @@ export const Button = React.forwardRef(
     ) : icon ? (
       <Icon
         icon={icon}
+        title={title}
```
barsukov, line 133: Would be possible to introduce the tests for this component as well?

barsukov (changes requested):
Good catch, I have one comment for this one.

packages/ui-components/src/components/Button/Button.component.js
VoigtS: I added a unit test for this case now.

edda: To ensure the displayed title matches the button title, you should use `titleValue` instead of the raw prop value.

```suggestion
        title={titleValue}
```

edda (changes requested):

packages/ui-components/src/components/Button/Button.component.js
VoigtS: Yes, you are right, the titleValue is preferable.

edda (approved):

packages/ui-components/src/components/Button/Button.test.js
```diff
@@ -131,4 +131,10 @@ describe("Button", () => {
     expect(screen.getByRole("button")).toHaveAttribute("id", "button-1")
     expect(screen.getByRole("button")).toHaveAttribute("data-lolol")
   })
+
+  test("renders an icon with the defined button title", async () => {
```
barsukov, line 135: Please remove async when you do not use await, lint going to throw error later on when we transfer it to TypeScript

barsukov (changes requested):
My last little comment

packages/ui-components/src/components/Button/Button.test.js
VoigtS: No problem. I was just following the given test structure, so be aware that most functions are async while not using an await. You might have to fix this later for the remaining ones.

barsukov (approved):

PR #302, opened 2024-09-04
fix(ui): add upload icon into icon lib

VoigtS:
# Summary

During my development duties I noticed that Juno offers a `Download` Icon, but no `Upload` Icon. I would like to include it.

- Add the Icon from the material-design lib
- Bump the juno version

# Screenshots (if applicable)
image

# Testing Instructions
Open the Storybook, navigate to the Icon section, View the `Upload` Icon


barsukov (changes requested):
If you would like to introduce new Icon to the Icon.js I would suggest to do the same for IconTs cause we are currently in a migration phase that try to migrate the whole components to TS for that reason please also implement this functionality in IconTS.component.ts

VoigtS:
> If you would like to introduce new Icon to the Icon.js I would suggest to do the same for IconTs cause we are currently in a migration phase that try to migrate the whole components to TS for that reason please also implement this functionality in IconTS.component.ts

Thank you for the info, I included the icon in the Ts implementation as well.

edda (approved):

barsukov (approved):

PR #306, opened 2024-09-05
chore(ui): transfer ToolTip to Typescript

barsukov:
# Summary

Finish the ToolTip migration to typescript.

# Changes Made

# Related Issues

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


edda (changes requested):
The tooltip positioning is off when the `asChild` option is used. See screenshot.
Screenshot 2024-09-05 at 16 39 52

edda (changes requested):
For the props where we have specific possible values, e.g. `variant` or `placement` the custom Type leads to the storybook documentation to state `union` as the type. Previously when we used PropTypes it would list all possible values, which I think makes much more sense. Can we get a similar behaviour back?
Screenshot 2024-09-05 at 16 49 31
Screenshot 2024-09-05 at 16 49 41

edda (approved):

edda (approved):

PR #324, opened 2024-09-10
chore(ui): introduce folder with deprecated JS components

gjaskiewicz:
# Summary

Move components which were migrated to TS, but still need JS version to component-transition folder.
Place all JS components which has TS components, but need to stay in "Deprecated/" tree hierarchy in storybook.

# Changes Made

Moved files and updated dependencies

# Testing Instructions

1. `npm i`
2. `turbo test`

# Checklist


packages/ui-components/src/components/Icon/Icon.component.tsx
```diff
@@ -21,7 +21,7 @@ import ChevronLeft from "@material-design-icons/svg/outlined/chevron_left.svg"
 import ChevronRight from "@material-design-icons/svg/outlined/chevron_right.svg"
 import Close from "@material-design-icons/svg/filled/close.svg"
 import ContentCopy from "@material-design-icons/svg/outlined/content_copy.svg"
-import Danger from "./../Icon/icons/juno-danger.svg"
+import Danger from "../../deprecated_js/Icon/icons/juno-danger.svg"
```
edda, line 24: What is the reason that you load these icons from the from the `deprecated_js` folder here in the TypeScript Icon component? They are custom icons we created ourselves. We don't want to get rid of them.

edda (changes requested):

packages/ui-components/src/components/Icon/Icon.component.tsx
gjaskiewicz: No, I just missed them out. Thank you for catching this. I have moved the svg folder to TS Icon component folder.

edda (approved):

andypf (approved):

PR #343, opened 2024-09-12
feat(supernova): enable editing name, if user is anonymous

TilmanHaupt:
# Summary

For all silence forms its now enabled to edit the username if a user is anonymous.

# Related Issues

closes https://github.com/cloudoperators/juno/issues/335

# Checklist


hodanoori (approved):
lgtm

apps/supernova/src/components/silences/RecreateSilence.jsx
```diff
@@ -82,6 +83,9 @@ const RecreateSilence = (props) => {
     setError(null)
     setSuccess(null)
     setShowValidation({})
+    if (authData?.parsed?.fullName === "anonymous") {
```
ArtieReus, line 86: this looks like too weak... I would suggest to use at least a constant which is used in the setter and here. But the best would be to set a flag in the AuthData object which defines that there is no authentication a sets default attributes. what do you think?

apps/supernova/src/components/silences/RecreateSilence.jsx
```diff
@@ -198,7 +202,13 @@ const RecreateSilence = (props) => {
 
               <Form className="mt-6">
                 <FormRow>
-                  <TextInput required label="Silenced by" value={formState.createdBy} disabled />
+                  <TextInput
+                    required
+                    label="Silenced by"
+                    value={formState.createdBy}
+                    disabled={!isNameEditable}
```
ArtieReus, line 209: the same here. You wouldn't need to check with a local variable if the authData is set correctly with a flag which describes that there is no authentication.

apps/supernova/src/components/silences/SilenceNew.jsx
```diff
@@ -78,6 +80,10 @@ const SilenceNew = ({ alert, size, variant }) => {
       createdBy: authData?.parsed?.fullName,
       matchers: setupMatchers(alert?.labels, excludedLabels, enrichedLabels),
     })
+
+    if (authData?.parsed?.fullName === "anonymous") {
```
ArtieReus, line 84: Same here. No need to set again this information in each component if it is set in the store.

apps/supernova/src/components/silences/SilenceScheduled.jsx
```diff
@@ -155,8 +157,18 @@ const SilenceScheduled = (props) => {
     setFormState(newFormState)
 
     setSelected(newSelectedOption)
+
+    if (authData?.parsed?.fullName === "anonymous") {
```
ArtieReus, line 161: same same

ArtieReus (changes requested):

apps/supernova/src/components/silences/SilenceScheduled.jsx
```diff
@@ -72,10 +78,10 @@ const SilenceScheduled = (props) => {
 
     if (errorFormState) {
       setFormState(errorFormState)
-      addMessage({
-        variant: "error",
-        text: parseError("Please fix the errors in the form"),
-      })
+      // addMessage({
```
TilmanHaupt, line 81: @ArtieReus

.changeset/cyan-weeks-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudoperators/juno-app-supernova": patch
```
edda, line 2: please turn this into a minor change. It's a new feature. Patch is only for bug fixes or small changes that don't introduce a new feature/behavior
```suggestion
"@cloudoperators/juno-app-supernova": minor
```

apps/supernova/src/components/silences/SilenceScheduled.jsx
```diff
@@ -247,7 +259,13 @@ const SilenceScheduled = (props) => {
             <Form>
               <FormSection>
                 <FormRow>
-                  <TextInput required label="Silenced by" value={formState.createdBy} disabled />
+                  <TextInput
+                    required
+                    label="Silenced by"
+                    value={formState.createdBy}
+                    onChange={(e) => onInputChanged({ key: "createdBy", value: e.target.value })}
```
edda, line 266: You should validate the input here and show an error state for the `TextInput` if the value is empty. 
Also the "Save" button should be disabled as long as the form has errors.

edda (changes requested):

apps/supernova/src/components/silences/SilenceScheduled.jsx
TilmanHaupt: already in work

apps/supernova/src/lib/createAuthDataSlice.js
```diff
@@ -12,19 +12,28 @@ const createAuthDataSlice = (set, get) => ({
...
+          anonymous = true
+        }
+
         set(
           (state) => ({
             auth: {
               ...state.auth,
               loggedIn: data?.loggedIn,
               data: data?.auth,
+              userEditable: !anonymous,
```
ArtieReus, line 36: why not doing this?
!!data.auth?.parsed?.fullName

apps/supernova/src/lib/createAuthDataSlice.js
```diff
@@ -12,19 +12,28 @@ const createAuthDataSlice = (set, get) => ({
...
     loggedIn: false,
+    userEditable: true,
 
     actions: {
       setData: (data) => {
         if (!data) return
         // check if data has changed before updating the state
         if (data?.loggedIn === get().auth.loggedIn && data?.auth === get().auth.data) return
 
+        // If a user has a name, name is not editable in forms.
```
ArtieReus, line 23: this can be removed. See below

apps/supernova/src/hooks/useCommunication.js
```diff
@@ -60,7 +60,6 @@ const useCommunication = () => {
   }, [setIsActive])
 
   useEffect(() => {
-    authSetData({ auth: { parsed: { fullName: "anonymous" } } })
```
ArtieReus, line 63: I am missing the new changes here

apps/supernova/src/components/silences/SilenceNew.jsx
```diff
@@ -75,9 +82,10 @@ const SilenceNew = ({ alert, size, variant }) => {
     setFormState({
       ...formState,
       ...DEFAULT_FORM_VALUES,
-      createdBy: authData?.parsed?.fullName,
+      createdBy: authData?.parsed?.fullName || "",
```
ArtieReus, line 85: this would be better to be done in the store or?

apps/supernova/src/components/silences/RecreateSilence.jsx
```diff
@@ -70,7 +76,7 @@ const RecreateSilence = (props) => {
     setFormState({
       ...formState,
       ...DEFAULT_FORM_VALUES,
-      createdBy: authData?.parsed?.fullName,
+      createdBy: authData?.parsed?.fullName || "",
```
ArtieReus, line 79: why you do this? and why not in the store?

apps/supernova/src/components/silences/RecreateSilence.jsx
```diff
@@ -27,11 +27,16 @@ import constants from "../../constants"
 
 const validateForm = (values) => {
   const invalidItems = {}
-  if (values?.comment?.length <= 3) {
+  if (values?.comment?.length < 3) {
     if (!invalidItems["comment"]) invalidItems["comment"] = []
     invalidItems["comment"].push(`Please enter at least 3 characters`)
   }
 
+  if (values?.createdBy?.length < 1) {
```
ArtieReus, line 35: should not be <= 0?

ArtieReus (changes requested):

ArtieReus (dismissed):
lgtm

ArtieReus:
Improvement proposal:
- When saving the Auth object just extract the relevant attributes like `fullname` and `support_group` and just save them in the Store. So you would save to check in different places `fullname || ""` when you initialize the fullname already with `""` in the store.

barsukov (commented):
LGTM

apps/supernova/src/components/silences/SilenceScheduled.jsx
```diff
@@ -164,7 +164,6 @@ const SilenceScheduled = (props) => {
   }
 
   const onInputChanged = ({ key, value }) => {
-    if (!value) return
     setFormState({ ...formState, [key]: { value: value, error: null } })
     console.log("formState", formState)
```
edda, line 168: Please remove the left over `console.log`

edda (changes requested):

apps/supernova/src/components/silences/SilenceScheduled.jsx
```diff
@@ -157,6 +163,12 @@ const SilenceScheduled = (props) => {
     setSelected(newSelectedOption)
   }
 
+  const onInputChanged = ({ key, value }) => {
+    setFormState({ ...formState, [key]: { value: value, error: null } })
+    console.log("formState", formState)
```
edda, line 168:
```suggestion

```

edda (changes requested):

edda (approved):

apps/supernova/src/components/silences/RecreateSilence.jsx
```diff
@@ -27,11 +27,16 @@ import constants from "../../constants"
 
 const validateForm = (values) => {
   const invalidItems = {}
-  if (values?.comment?.length <= 3) {
+  if (values?.comment?.length < 3) {
```
barsukov, line 30: Just a little comment, would be nice to extract this magic number into a constant please have look here https://stackoverflow.com/questions/47882/what-are-magic-numbers-and-why-do-some-consider-them-bad

TilmanHaupt: I do quite often constants. And not do it in a validation form where I just fixed the number that is equal to the error message. "please add at least 3 characters". Its also not global.

hodanoori (dismissed):
lgtm

barsukov (approved):

andypf (approved):

PR #352, opened 2024-09-13
feat(ui): fixes Pagination inconsistencies. Adds disabled and loading state

### LINKED ISSUE #328: [Task](ui): Pagination fixes
**Task Description**
Improvements for the Pagination component. There are various issues with the `Pagination` component due to the fact that internally it uses an uncontrolled `Select` in the select variant. This leads to unexpected behaviour when the initial props that are passed to the component change (e.g. because the total number of pages was retrieved asynchronously after the component was already rendered. To fix this we will use the Select in controlled mode and keep internal state in the `Pagination` component itself. This way it can by used like an uncontrolled or a controlled component. This will lead to less confusion for app developers.

Additionally add some extra features to the Pagination component: disabled state for the whole component, loading state
Bug fix: input field width should be dynamic in case of huge numbers

**Sub-tasks**

- [x] add new prop `disabled` to disable whole component
- [x] add new prop `loading` that shows loading state
- [x] change internal Select to use `value` instead of `defaultValue` so that we use it in controlled mode
- [x] add internal state (currentPage, pages)
- [x] add useEffect for props to make it possible to update Pagination
- [x] dynamic input field width

Fixes:
- [x] Do not set default value of `1` to currentPage and totalPages. Instead keep them undefined. In all event handlers handle undefined for these. Ensure currentPage changes only happen if currentPage is not falsy.
- [x] If `currentPage` is set prev button always decreases currentPage by 1 (1 is minimum), next button always increases currentPage by 1 (if totalPages is set, the following is true: `1 <= currentPage <= totalPages)
- [x] Event handlers that change the value of currentPage should return the new value, otherwise undefined
- [x] Add tests for all the edge cases

MartinS-git:
# Summary

This is a recreation of a PR that got left behind in the old repo. For details see PR: https://github.com/sapcc/juno/pull/564 
The old PR have been transferred to the current status of the JunoUI components. 
Minor bug fixes due to the lint validation have been done.

Changes:
- add new prop `disabled` to disable whole component
- add new prop `loading` that shows loading state
- change internal Select to use value instead of defaultValue so that we use it in controlled mode
- add internal state (currentPage, pages)
- add useEffect for props to make it possible to update Pagination
- dynamic input field width

Closes #328 

# Checklist


packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -3,117 +3,218 @@
...
+  onBlur = undefined,
+  progress = undefined,
   className = "",
   ...props
 }) => {
+  const [controlPage, setControlCurrentPage] = useState(1)
+  const [controlTotalPage, setControlTotalPage] = useState(1)
+
+  useEffect(() => {
+    setControlCurrentPage(currentPage)
```
barsukov, line 49: I would like to discuss if it does make sense to introduce use effect here that just control the state and no side effects. What do you think that we could have a call and talk about it :)

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -3,117 +3,218 @@
...
+        icon="chevronRight"
+        disabled={isLastPage || disabled || progress || controlPage === controlTotalPage}
+        onClick={handleNextClick}
+        title="Next Page"
+      />
     </div>
   )
 }
-
 Pagination.propTypes = {
```
barsukov, line 175: it looks like almost a new component for me, I would suggest to try to write in Typescript instead

packages/ui-components/src/components/Pagination/Pagination.test.js
```diff
@@ -87,6 +86,114 @@ describe("Pagination", () => {
...
+
+  test("renders Pagination (input) - fires onChange handler as passed", async () => {
+    const onChangeMock = jest.fn()
+    render(<Pagination variant="input" pages={12} onChange={onChangeMock} />)
+    const textinput = screen.getByRole("textbox")
+    fireEvent.change(textinput, { target: { value: "102" } })
+    expect(onChangeMock).toHaveBeenCalledTimes(1)
+  })
+
+  test("renders Pagination (input) - check whether a value that is too high is automatically corrected to the highest possible value - as passed", async () => {
```
barsukov, line 134: Please remove keyword async everywhere where it does not needed in tests.

packages/ui-components/src/components/Pagination/Pagination.test.js
```diff
@@ -87,6 +86,114 @@ describe("Pagination", () => {
...
+    render(<Pagination variant="input" pages={12} />)
+    const textinput = screen.getByRole("textbox")
+    fireEvent.change(textinput, { target: { value: "102" } })
+    fireEvent.keyPress(textinput, { key: "Enter", code: 13, charCode: 13 })
+    expect(textinput).toHaveValue("12")
+  })
+
+  test("renders Pagination (input) - check whether a value that is too low is automatically corrected to the value : 1 - as passed", async () => {
```
barsukov, line 142: If you feel that description is too long, maybe it is a good idea to split the tests in group and add additional describe on it.

barsukov (changes requested):
Good work! I have few comments that would be easy to discsuss in a call :)

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -3,117 +3,218 @@
...
     <div
       className={`juno-pagination juno-pagination-${variant || "default"} ${paginationStyles} ${className}`}
       {...props}
     >
-      <Button icon="chevronLeft" disabled={isFirstPage} onClick={handlePrevClick} title="Previous Page" />
-      {variant ? renderPaginationInnards(variant, currentPage, pages, handleSelectChange, handleKeyPress) : null}
-      <Button icon="chevronRight" disabled={isLastPage} onClick={handleNextClick} title="Next Page" />
+      <Button
+        icon="chevronLeft"
+        disabled={isFirstPage || disabled || progress || controlPage === 1}
```
edda, line 110: Something in this boolean expression leads to this button to be disabled when rendering the default variant (i.e. only render the prev/next buttons). My guess is it's the `controlPage === 1` because the initial value for `controlPage` is `1`. You probably need another expression in here that ensures button isn't accidentally disabled in the "default" variant.

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -3,117 +3,218 @@
...
+                  </Stack>
+                )
+              default:
+                return ""
+            }
+          })()
+        : ""}
+      <Button
+        icon="chevronRight"
+        disabled={isLastPage || disabled || progress || controlPage === controlTotalPage}
```
edda, line 168: Same thing as the "prev" button. This button is disabled in the "default" variant even though it shouldn't be

edda (changes requested):

edda (changes requested):
Last comment: The PR title has changed to one of the commit messages. Not sure how this happened, but the PR title should state what this PR does instead.

packages/ui-components/src/components/Pagination/Pagination.test.js
MartinS-git: Description has been shortened

packages/ui-components/src/components/Pagination/Pagination.stories.js
```diff
@@ -28,22 +28,26 @@ export default {
...
-Template.propTypes = {
+UncontrolledTemplate.propTypes = {
   currentPage: PropTypes.number,
   pages: PropTypes.number,
 }
 
 export const Default = {
   render: Template,
-  args: {},
+  args: { variant: "default" },
```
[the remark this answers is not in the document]
MartinS-git: It only seems to be disabled because the values “currentPage” and “pages” are not set. As soon as you can neither click forward nor back, the buttons are in the disabled state. To correct this, I have added the above value.

packages/ui-components/src/components/Pagination/Pagination.component.js
MartinS-git: I don't quite understand. Has it perhaps been resolved with the previous comment?

packages/ui-components/src/components/Pagination/Pagination.component.js
MartinS-git: Also here?

packages/ui-components/src/components/Pagination/Pagination.component.js
barsukov: @MartinS-git let us discuss this thing and how we can fix it and test it.

barsukov (commented):
one little comment

packages/ui-components/src/components/Pagination/Pagination.component.js
MartinS-git: @barsukov Thank you. However, I still need to make a few changes to this component first.

barsukov (dismissed):
As we spoke we will do typescript migration afterwards

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -58,12 +60,15 @@ export const Pagination = ({
...
-    setControlCurrentPage(controlPage > 1 ? controlPage - 1 : 1)
+    if (controlTotalPage || pages || controlPage) {
+      setControlCurrentPage(controlPage > 1 ? controlPage - 1 : undefined)
+    }
     onPressPrevious && onPressPrevious(event)
   }
 
   const handleNextClick = (event) => {
-    setControlCurrentPage(controlPage < controlTotalPage ? controlPage + 1 : controlTotalPage)
+    setControlCurrentPage(controlPage < controlTotalPage || controlPage ? controlPage + 1 : controlPage)
```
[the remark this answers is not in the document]
edda: After having fixed this, please also add a test that checks that the buttons behave correctly also for edge cases (all combination of defined/undefined, too low, too high, ...)

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -3,117 +3,207 @@
...
+  variant: PropTypes.oneOf(["default", "number", "select", "input"]),
+  /** The current page number */
   currentPage: PropTypes.number,
+  /** The total number of pages */
+  totalPages: PropTypes.number,
+  /** The total number of pages (fallback solution, please use totalPages) */
   pages: PropTypes.number,
+  /** Disable component */
+  disabled: PropTypes.bool,
+  /** Is the first page (mostly "1") - left button disabled */
```
edda, line 191: Rephrase proposal
```suggestion
  /** Use this to have the Pagination emulate the state as being on the first page. Useful for when you don't have knowledge about pages but know that there is no previous page. Setting this to true leads to the left/prev button getting disabled */
```

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -3,117 +3,207 @@
...
   isFirstPage: PropTypes.bool,
+  /** Is the last page (e.g. total number of pages) - right button disabled */
```
edda, line 193: Rephrase proposal
```suggestion
  /** Use this to have the Pagination emulate the state as being on the last page. Useful for when you don't have knowledge about pages/total page count but know that there is no next page. Setting this to true leads to the right/next button getting disabled */
```

edda (changes requested):

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -75,9 +80,9 @@ export const Pagination = ({
 
   const handleEnter = (event) => {
     if (event.key === "Enter") {
-      if (controlPage < 1) {
+      if (controlPage && controlPage < 1) {
         setControlCurrentPage(1)
-      } else if (controlPage > controlTotalPage) {
+      } else if (controlPage && controlPage > controlTotalPage) {
```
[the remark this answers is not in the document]
MartinS-git: I have created a test and also checked it manually. No errors appear.

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -60,15 +60,18 @@ export const Pagination = ({
...
     }
+
     onPressPrevious && onPressPrevious(event)
   }
 
   const handleNextClick = (event) => {
-    setControlCurrentPage(controlPage < controlTotalPage || controlPage ? controlPage + 1 : controlPage)
-
+    // set controlPage +1 if controlPage is not undefined and controlPage is less than controlTotalPage. Set conrolPage +1 also if controlTotalPage is undefined.
+    if (!controlTotalPage || (controlPage && controlPage < controlTotalPage)) {
```
edda, line 72: If `controlPage` and `controlTotalPage` are both falsy then this expression evaluates to `true` and you will try to increment the value of `controlPage` even though you shouldn't. The reason for this is that as soon as one segment of a boolean expression connected via `||` is true then javascript stops evaluating further and will treat the expression as true.

You'll need to change this so it really only does anything if `controlPage` is truthy. Something like:
```suggestion
    if (controlPage) {
      if ((controlPage < controlTotalPage) || !controlTotalPage) {
```
I would suggest separating it into two nested ifs for readability.  Also please add a test that tests this edge case (controlPage truthy, controlTotalPage falsy)

packages/ui-components/src/components/Pagination/Pagination.test.js
```diff
@@ -84,6 +100,96 @@ describe("Pagination", () => {
...
+      />
+    )
+    act(() => {
+      screen.getByRole("button", { name: "Next Page" }).click()
+    })
+    expect(handlePressNext).toHaveBeenCalledTimes(1)
+    expect(screen.getByTestId("my-pagination")).toHaveTextContent("6")
+  })
+
+  test("fires onPressNext handler with higher currentPage than totalPages", () => {
```
edda, line 157: This test checks that the onPressNext handler is not fired in this case, so the test description should reflect this.
```suggestion
  test("does not fire onPressNext handler with higher currentPage than totalPages", () => {
```

packages/ui-components/src/components/Pagination/Pagination.test.js
```diff
@@ -110,6 +216,17 @@ describe("Pagination", () => {
     })
   })
 
+  test("fires onKeyPress handler on Enter as passed for input variant with undefined controlPage", async () => {
```
edda, line 219: same again, test checks that handler is not called in this case.
```suggestion
  test("does not fire onKeyPress handler on Enter for input variant with undefined controlPage", async () => {
```

edda (changes requested):

packages/ui-components/src/components/Pagination/Pagination.component.js
```diff
@@ -69,7 +69,8 @@ export const Pagination = ({
 
   const handleNextClick = (event) => {
     // set controlPage +1 if controlPage is not undefined and controlPage is less than controlTotalPage. Set conrolPage +1 also if controlTotalPage is undefined.
-    if (!controlTotalPage || (controlPage && controlPage < controlTotalPage)) {
+    if (controlPage) {
+      if ((controlPage < controlTotalPage) || !controlTotalPage) {
       setControlCurrentPage(controlPage + 1)
```
edda, line 74: The closing `}` is missing here.
```suggestion
        setControlCurrentPage(controlPage + 1)
      }
```

edda (changes requested):

packages/ui-components/src/components/Pagination/Pagination.test.js
```diff
@@ -114,7 +114,7 @@ describe("Pagination", () => {
     act(() => {
       screen.getByRole("button", { name: "Next Page" }).click()
     })
-    expect(handlePressNext).toHaveBeenCalledTimes(1)
+    expect(handlePressNext).toHaveBeenCalledTimes(0)
```
edda, line 117: this is wrong. the pressNext handler must always fire. You probably changed this because the test was failing after the above change where you accidentally moved the nextPress handler into the if so it didn't fire when controlPage was falsy. But that is not the intended behaviour. It must always fire.

When a test suddenly fails, first check why it's failing rather than changing the test

packages/ui-components/src/components/Pagination/Pagination.test.js
```diff
@@ -132,7 +132,7 @@ describe("Pagination", () => {
...
```
edda, line 135: see above

edda (changes requested):

barsukov:
This PR has conflicts and also it looks like @edda suggestions that made by github as separated commits did not start the checks, as I think there is something to do with permissions that we have on a repo.

MartinS-git (commented):
Discussed point modified with new commit

edda (approved):

edda (approved):
yay!

barsukov (approved):

barsukov (approved):

barsukov (approved):

edda (approved):

Issue #169, opened 2024-08-02
[Feature](ui): PageHeader - Allow custom brand logo in


edda:
**Description**
The `PageHeader` component should allow customization of the brand logo it renders.

**Proposed Solution**
Allow passing via prop a custom brand logo

MartinS-git:
Replaced SAP logo and Favicon in Storybook and in PageHeader component with PR: https://github.com/cloudoperators/juno/pull/304

PR #358, opened 2024-09-13
feat(ui): pageheader custom logo

franzheidl:
feat(ui): Pageheader custom logo

Allow users to pass a custom logo component to the PageHeader and render it. 

# Changes Made

- use grid layout in header instead of flexbox as it gives us better control to make sure the logo is scaled correctly while keeping its original aspect ratio
- add a `logo` prop that accepts a component to render in a slot
- render the default Juno logo if `true` or `undefined`
- do not render any logo at all if `false` or `null` 

# Related Issues

- #169  
# Screenshots (if applicable)

# Testing Instructions

1. `npm i`
2. `npm run test` / `npm run test PageHeader`

# Checklist


barsukov (changes requested):
I would suggest to close this PR for now cause @gjaskiewicz works on transition this component to Typescript in current sprint, to avoid future conflicts, if it is urgent then maybe we could add this changes to his PR. The task is #212

PR #395, opened 2024-09-19
fix(supernova): bind disabled TextInput directly to authName

TilmanHaupt:
# Summary

Bind disabled TextInput directly to authName to ensure correctness. To ensure the field is disabled if there is a name.


ArtieReus (approved):
lgtm

andypf (approved):

edda (approved):

PR #405, opened 2024-09-22
chore(example): switch to vite and vitest 

andypf:
# Summary

This PR replaces esbuild with Vite and Jest with Vitest, unifying the build and testing processes across the project. This transition eliminates many dependencies, such as Babel, resulting in a leaner setup while preparing the codebase for future TypeScript integration.

With Vite, we gain the ability to generate static builds that can easily be deployed to platforms like GitHub Pages. Vite uses an index.html file located in the project’s root as the entry point, then generates a final index.html with all asset references dynamically replaced by URLs pointing to the optimized, generated assets.

A key advantage of Vite is that it supports this static build workflow even during development, meaning you can preview the final production-like setup with live-reloading and asset management in real-time, streamlining the development process and improving efficiency.

1. `npm i`
2. `npm run test`

# Checklist


.husky/pre-commit
```diff
@@ -7,5 +7,5 @@
 # Now you can use npm or any Node.js commands
 
 # npm run lint
-npm run lint -- -- --fix
-npm run format
+#npm run lint -- -- --fix
```
barsukov, line 10: I just leave a smile here :) if you need to ignore husky next time, please do it with env variable instead

.husky/pre-push
```diff
@@ -3,4 +3,4 @@
 # .husky/pre-push
 
 # Run lint, typecheck, and test using Turbo
-npx turbo run lint typecheck --concurrency=50%
+#npx turbo run lint typecheck --concurrency=50%
```
barsukov, line 6: same here

apps/example/vite.config.ts
```diff
@@ -0,0 +1,66 @@
...
+        plugins: [tailwindcss, autoprefixer],
+      },
+    },
+
+    server: {
+      host: "0.0.0.0",
+      port: parseInt(process.env.PORT || "3000"),
+    },
+
+    /* enable jsx in js files */
```
barsukov, line 26: Please delete commented code

barsukov (changes requested):
Looks good please rollback husky

.husky/pre-commit
andypf: 👍

.husky/pre-push
andypf: 👍

apps/example/vite.config.ts
andypf: :saluting_face:

barsukov (dismissed):
LGTM

ArtieReus (approved):
lgtm just look at the commet I added

apps/example/index.html
```diff
@@ -3,11 +3,11 @@
   ~ SPDX-License-Identifier: Apache-2.0
 -->
 
-<!DOCTYPE html>
+<!doctype html>
```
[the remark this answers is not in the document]
andypf: good point, lowercase is outdated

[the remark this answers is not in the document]
andypf: in HTML, the DOCTYPE is case insensitive! What ever you like :)

hodanoori (approved):

apps/example/README.md
```diff
@@ -3,68 +3,60 @@
...
 ```
 
-Or using React's lazy loading to keep the bundle size small
+## Development Mode
 
-```js
-import { lazy } from "react"
+```bash
+cd apps/example
+npx turbo dev
```
edda, line 46: Shouldn't we instead refer to the npm script here (in case we decide to use something other than turbo in the future)?
```suggestion
npm run dev
```

apps/example/README.md
```diff
@@ -3,68 +3,60 @@
...
-  /*...*/
-  return (
-    <div>
-      <Exampleapp />
-    </div>
-  )
-}
+```bash
+cd apps/example
+npx turbo test
```
edda, line 53: Ditto
```suggestion
npm run test
```

edda (changes requested):

apps/example/README.md
andypf: currently we use turbo to build dependencies for this app. Without turbo it will not work.

apps/example/README.md
andypf: dito :)

edda (approved):

PR #448, opened 2024-09-24
fix(ui): fix placement for ContextMenu content

VoigtS:
# Summary

With further implementation of the Roadmap features for resource commitments I came to the conclusion that I need to group the operations that a user can perform on a commitment.
As a solution I picked the ContextMenu. However, I quickly noticed it suffers from the same issue the `Select` function had. 
If no space is available, the children list overflows, making it unaccessible for the user.

# Changes Made

- Include the functionality of floating-ui within the ContextMenu in order to prevent MenuItem overflows.
- Add cases to test on in the storybook
- "Fix" a broken import from the MenuItems which caused the Storybook to error out.
(Not a true fix. It just does the same how it was handled in the `Breadcrum` component and reverts the import back to the old one.)

# Related Issues

A solution for the `Select` item was provided in PR #204. This PR aims to fix the issue in a similar way.
For that, the flip was utitilized in order to ensure a default placement (the default is bottom-start - the same layout that the original implementation inherited.). This avoids an inconsistent placement with `autoPlacement` which would result in a left or right placement depending on which side has more space available.

# Screenshots (if applicable)
Example of the fix:
image

# Testing Instructions

- Open the Storybook
- Navigate to `ContextMenu` and scroll down
- View the ContextMenu examples.

I did not implement additional unit tests, because the placements are not fit to be included in them. Therefore the Storybook is the main source to test out the component fix (the linked PR handles the case similarly).


packages/ui-components/src/components/ContextMenu/ContextMenu.component.js
```diff
@@ -60,7 +61,7 @@ export const ContextMenu = ({
 
   return (
     <Menu>
-      <Float>
+      <Float middleware={[flip()]}>
```
barsukov, line 64: I follow a simple rule:
You touched the functionality of a component(change behaviour, extend or delete).
You add the test :)

packages/ui-components/src/components/MenuItem/MenuItem.stories.js
```diff
@@ -7,7 +7,7 @@ import React from "react"
 import { MenuItem } from "./MenuItem.component"
 import { Menu } from "../Menu/Menu.component"
 import { Button } from "../Button/index"
-import { knownIcons } from "../Icon/Icon.component"
+import { knownIcons } from "../../deprecated_js/Icon/Icon.component.js"
```
barsukov, line 10: 👍

barsukov (changes requested):
@edda Please have a look on a necessity of this extension of this component 🙏 
From my side a little test is missing.

packages/ui-components/src/components/ContextMenu/ContextMenu.component.js
VoigtS: If you have an idea how to properly wrap the positioning check of the subcomponent into a unit test check, I'm open to implement one. In the PR comment I referred to the previous PR not having a dedicated unit test for its case either.

edda: You reference the fix from https://github.com/cloudoperators/juno/pull/204 
However there to fix the issue we removed the flip middleware and replaced it with autoPlacement.  I recommend you do the same here so that they behave the same.

edda (changes requested):

packages/ui-components/src/components/ContextMenu/ContextMenu.component.js
VoigtS: I would rather request the opposite. That I change the `autoPlacment` in that component to `flip` as well.
Because I noticed a bug by switching to public (npm) juno.
Autoplacement aligns the subcomponent to the area that has more space left. This means that the `Select` component is aligned downwards if there is more space or upwards, if there is more space left.

You can see it in the following example:
image

The contents are aligned upwards, because there is more screenspace left, not because we are out of space.

edda: yes, this was described in the referenced PR as well. `flip`, if it works in all cases, would definitely be preferred behaviour but as I mentioned, there were edge cases where it didn't work as expected. autoPlacement is not ideal and I would prefer flip but then we'll have to fix those edge cases. 

We are currently working on a fix to another esoteric placement issue with the DateTimePicker. Maybe we're lucky and the fix there also fixes the placement problems for the flip middleware. We'll have a look.

VoigtS: Thank you for you insight, I wasn't fully aware of the edge cases here. I will include the `autoPlacement` config.

VoigtS: The config now matches the `Select` counterpart. I had to change the allowed placements to match the desired behavior of the ContextMenu specifically tho.

There is also an issue with the overflow wich is present in the `Select` component - if you scroll down enough not all items are accessible in within the scroll list. This changes the further up you scroll in the documentation. This might be an issue originating from the box Storybooks box resizing.
This is now present in the ContextMenu as well. However if you look at the placement issue anyways, it is as you said, better to have matching behaviors on both components.

edda (approved):
I'll approve this so you aren't blocked. However please note that the ContextMenu component is still in the "WIP" which means that it's either not finished or might get refactored. I this case we will most likely completely refactor it. So pay attention to warnings about deprecations of the current component in some future release.

barsukov (approved):

barsukov (approved):

barsukov (approved):

Issue #456, opened 2024-09-25
[Task](ci): preview example app when PR contains changes in ui-components 


ArtieReus:
**Task Description**
Preview example app when PR contains changes in ui-components.

**Sub-tasks**

- [x] Change workflow to preview all apps when ui-changes have been done

**Related Issues**
Link to related issues or tasks, if any.

**Additional Context**
Add any other context or information here.

PR #472, opened 2024-09-27
chore(ci): fails preview workflow if any build fails. Display failed build on the index.html

ArtieReus:
# Summary

Fail preview workflow if any build fails. Display failed build on the preview index.html

# Changes Made

- Added example app
- Changes to ui-components in a PR now generate previews for both Storybook and Example.
- The Vite compile step has been moved to a bash script.
- Failed previews are now displayed on the index.html page.
- Links to the previews open in a new tab.
- The index.html was refactored with a loop to reduce code

# Screenshots (if applicable)

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


.github/workflows/deploy-pr-preview.yaml
```diff
@@ -124,18 +128,34 @@ jobs:
           HEUREKA_CHANGED: ${{ contains(needs.run-detect-changes.outputs.changes, 'heureka') }}
           DOOP_CHANGED: ${{ contains(needs.run-detect-changes.outputs.changes, 'doop') }}
         run: |
```
andypf, line 130: You can simplify this code by using an array to handle the components dynamically. This approach eliminates duplication and makes it much easier to add new apps in the future:

```yaml
run: |
  COMPONENTS=("ui-components" "supernova" "heureka" "doop")

  echo "<!DOCTYPE html><head><title>PR Preview</title></head><style> .failed { color: red; font-weight: bold; } </style><html><body><h1>Deployed Packages for PR ${GITHUB_HEAD_REF}</h1><ul>" > ${{ env.DEPLOY_PATH }}/index.html

  for component in "${COMPONENTS[@]}"; do
    CHANGED=$(echo "${{ contains(needs.run-detect-changes.outputs.changes, component) }}")
    BUILD_STEP="build-${component}"

    if [ "$CHANGED" = "true" ]; then
      if [ "${{ steps.${BUILD_STEP}.outcome }}" == "success" ]; then
        echo "<li><a href='./${component}/index.html'>${component^} (Success)</a></li>" >> ${{ env.DEPLOY_PATH }}/index.html
      else
        echo "<li class='failed'>${component^} (Failed)</li>" >> ${{ env.DEPLOY_PATH }}/index.html
      fi
    fi
  done

  echo "</ul></body></html>" >> ${{ env.DEPLOY_PATH }}/index.html
  ```

andypf (commented):
I have a suggestion to simplify the code and make it more maintainable—let me know what you think!

.github/workflows/deploy-pr-preview.yaml
ArtieReus: oh! let me have a look. I didn't thought about it

barsukov (commented):
Just a question why could we have one page html one and put all logic there instead of try to write it in a yaml file which makes it unmaintainable, have you check how others do it ? Maybe there are some common way?

ArtieReus:
> Just a question why could we have one page html one and put all logic there instead of try to write it in a yaml file which makes it unmaintainable, have you check how others do it ? Maybe there are some common way?

The logic to check which apps have changed and if the application can be build for the preview can just be done in the workflow. The page builds on this parameters.

ArtieReus:
Fixes: #456

barsukov (commented):
I leave it here, nice refactoring it became really better but I still think maybe we could extract this page if possible into html and make it stylish but not inside the yml... not sure how we could do that, but we should also have a look how others solve this problem

ArtieReus:
> I leave it here, nice refactoring it became really better but I still think maybe we could extract this page if possible into html and make it stylish but not inside the yml... not sure how we could do that, but we should also have a look how others solve this problem

this could be an improvement. Lets toggle it please in another TASK/PR.

andypf (approved):
LGTM

ArtieReus:
> I leave it here, nice refactoring it became really better but I still think maybe we could extract this page if possible into html and make it stylish but not inside the yml... not sure how we could do that, but we should also have a look how others solve this problem

Agreed with @barsukov to create a Proposal with a better solution.

hodanoori (approved):
lgtm

PR #484, opened 2024-09-30
fix(supernova): errors from the silencesAPI are just parsed if they are not string

### LINKED ISSUE #474: [Bug](supernova): Runtime error when faulty alertmanager endpoint is configured
**Description**
When something is wrong with the alertmanager endpoint then there is a runtime error in Supernova relating to the Messages component.

```
Objects are not valid as a React child (found: [object Error]). If you meant to render a collection of children, use an array instead.
```

**Steps to Reproduce**

1. configure a faulty alertmanager enpoint for Supernova in dev mode
2. start Supernova in dev 
3. See error

**Expected Behavior**
The error needs to be caught and displayed properly

**Screenshots**
Image

TilmanHaupt:
fixes #474 

# Changes Made

Checks if error message is a string and not a object and does not parse if its a string.

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist

### DIFF
--- a/apps/supernova/src/AppContent.jsx
+++ b/apps/supernova/src/AppContent.jsx
@@ -40,11 +40,13 @@ const AppContent = () => {
   const activeSelectedTab = useGlobalsActiveSelectedTab()
 
   useEffect(() => {
+    const errorMessage = typeof silencesError === "string" ? silencesError : parseError(silencesError)
+
     // since the API call is done in a web worker and not logging aware, we need to show the error just in case the user is logged in
     if (!silencesError) return
     addMessage({
       variant: "error",
-      text: parseError(silencesError),
+      text: errorMessage,
     })
   }, [silencesError])
 
--- a/apps/supernova/src/components/silences/SilenceScheduled.jsx
+++ b/apps/supernova/src/components/silences/SilenceScheduled.jsx
@@ -125,7 +125,7 @@ const SilenceScheduled = (props) => {
       .catch((error) => {
         addMessage({
           variant: "error",
-          text: parseError(error),
+          text: `${parseError(error)}`,
         })
       })
   }


edda (approved):

apps/supernova/src/AppContent.jsx
```diff
@@ -40,11 +40,13 @@ const AppContent = () => {
   const activeSelectedTab = useGlobalsActiveSelectedTab()
 
   useEffect(() => {
+    const errorMessage = typeof silencesError === "string" ? silencesError : parseError(silencesError)
+
     // since the API call is done in a web worker and not logging aware, we need to show the error just in case the user is logged in
     if (!silencesError) return
     addMessage({
       variant: "error",
-      text: parseError(silencesError),
```
barsukov, line 47: maybe you could modify the function itself instead of changing it this way ?

barsukov (changes requested):
one comment from my side :)

apps/supernova/src/AppContent.jsx
TilmanHaupt: +

barsukov (dismissed):
Good job, from my side is ok, please ask second reviewer :)

apps/supernova/src/helpers.js
```diff
@@ -4,37 +4,34 @@
  */
 
 export const parseError = (error) => {
-  let errMsg = error
-
   // check if error is JSON containing message or just string
   if (typeof error === "string") {
-    errMsg = parseMessage(error)
+    if (error === "Failed to fetch") {
```
edda, line 9: I recommend not doing a hard equality check here but a more soft substring check. It's more robust if the error message changes slightly.
```suggestion
    if (error.includes("Failed to fetch")) {
```

edda (changes requested):
I like the new changes +1 
Just one small suggestion to make it a bit more robust

edda (approved):

barsukov (approved):
LGTM

PR #488, opened 2024-10-01
feat(supernova): remove auth and pass initial filters per appProp

TilmanHaupt:
# Summary

This PR removes the authentication integration from the application, simplifying the access process and allowing a full preview of the application within a PR. Additionally, it introduces a new appProp called initialFilters and username, which allows default filters and auditing to be applied when initializing the application. 

# Related Issues

#368 
closes #422 

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [ ] I have made corresponding changes to the documentation (if applicable).
- [x] My changes generate no new warnings or errors.

PR #489, opened 2024-10-01
chore(greenhouse): adds temporary fix to forward default initial filters to the applications

ArtieReus:
# Summary

This PR is a FOLLOW-UP to the removal of authentication in Supernova and Doop. It adds a TEMPORARY fix to forward default initial filters to the applications until we come up with the new design. Specifically, it passes the `initialFilters` prop, containing `support_group` and `user`, to all applications to compensate for the missing authentication. There is already another task to redesign this process.

This PR is NOT risky; it simply passes additional props to the applications. See the debug output from Greenhouse:

```
{
   ...
    "endpoint": "https://<secret-URL>/api/v2",
   ...
    "username": "A Reuschenbach Puncernau",
    "initialFilters": {
      "support_group": ["containers"],
      }
    "id": "alerts-ccloud"
}
```

# Changes Made

- Add initialFilters with the support group and user full name

# Related Issues

- Issue 1: [link to issue]

# Screenshots (if applicable)

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


TilmanHaupt (changes requested):
Why don't we use the same format in initialFilters like we already use in my PR and activeFilters? Also we should instantly add the username (auth.parsed.fullName) to allow to delete auth like in https://github.com/cloudoperators/juno/pull/488.  || Prop examples:   initialFilters: {
    group: ["groupA", "groupB"],
  },
  username: "Franz Kafka",

TilmanHaupt (approved):

andypf:
This is really problematic. Several things are happening here:

First, in my opinion, the placement of the logic in useApi is not correct. It assumes that the auth object is already available, which happens to be the case here due to external circumstances. I would suggest moving this to Plugin.jsx, where both auth and config are definitely available.

Secondly, the `initialFilters` property is being passed to every app. But what’s particularly critical here is the `if` condition. If someone decides to change the name in PluginConfig, this will break. I understand that this is a temporary solution, but how is it supposed to work in the future?

There needs to be a piece of information that both sides know and that each app can process. My suggestion would be to pass an additional prop to every app, which could be called `auth` or `user`. Doop and Supernova, for example, could check if initialFilters is set. If not, they could then check if the `auth` property exists and try to read the filters from it. The auth object **shouldn’t** just be passed directly from OAuth to the apps, but should be **reconstructed** within Greenhouse. This would then form the interface between Greenhouse and the extensions, allowing each extension to access this information.

ArtieReus:
Thanks @andypf for the analysis! As mentioned, this is a temporary fix to unblock two other PRs and was agreed upon with @edda. Whether it belongs in the hook or the component doesn’t matter to me, as it will need to be redesigned and made configurable for Greenhouse in the future.

ArtieReus:
The plugins should be authentication-agnostic, meaning they don’t need to handle authentication directly—just the `initialFilters` (and the username for Supernova). We shouldn't push the logic into the plugins; instead, it should remain within the parent applications, like Greenhouse. How this will work in the future will depend on how we design the carbon template.

ArtieReus:
We have already a task to taggle the new design: https://github.com/cloudoperators/juno/issues/492

andypf (approved):

TilmanHaupt (approved):

Issue #368, opened 2024-09-02
[Task](supernova): Make initialFilters configurable as props


edda:
**Task Description**
Supernova allows the setting of initial filters. Currently this is only used for setting the user's support group as an initial filter. This needs to be generalized so that arbitrary filters can be initially active. These filters must be passable to Supernova via prop and readable from a config in the case of standalone deployment.

**Sub-tasks**

- [ ] Read initial filters from props. 
  - Variable name: `initialFilters`
  - Example:
  ```
   "initialFilters": {
     "support_group": ["containers"],
    }
  ```
- [ ] Read user name from props
  - Variable name: `username`
  - Example:
  ```
   "username": "A Reuschenbach Puncernau"
  ```
- [ ] Make initial filters configurable from env in standalone deployment

**Related Issues**
Link to related issues or tasks, if any.

Related Epic
#275

PR #500, opened 2024-10-08
fix(supernova): filters invalid keys from initialFilters

edda:
This PR changes the behaviour of setting the initial filters in the case that the given inital filters contain invalid filter labels (invalid meaning: labels that aren't defined as valid filter labels). Previously an empty object was returned. Now only the invalid labels are filtered out and the valid ones are kept.

# Summary

The old solution meant that if in a list of five initial filters one wasn't valid then no initial filters were set at all. Now the four valid ones will be set as expected and the invalid one will be ignored in addition to emitting the warning on the console which explains why this happened.

# Changes Made

- Added filter to the given initialFilters object which filters out invalid keys
- Added a test that verifies that the invalid key filter works

# Related Issues

- #368 

# Checklist


barsukov (commented):
I did not test it but LGTM from code wise

andypf (approved):

apps/supernova/src/lib/createFiltersSlice.test.js
```diff
@@ -288,18 +288,23 @@ describe("createFiltersSlice", () => {
...
+      const store = renderHook(
         () => ({
           activeFilters: useActiveFilters(),
         }),
         { wrapper }
       )
 
       expect(spy).toHaveBeenCalledWith(
-        "[supernova]::parseInitialFilters: Some keys of the initialFilters object are not in the labels"
+        "[supernova]::parseInitialFilters: Some keys of the initialFilters object are not valid filter labels. They must be configured as filterLabels first. Using only valid keys."
```
ArtieReus, line 305:
```suggestion
        "[supernova]::parseInitialFilters: Some keys of the initialFilters object are not valid filter labels. They must be configured as filterLabels first. Using only valid keys."
        expect.stringContaining("[supernova]::parseInitialFilters: Some keys of the initialFilters object are not valid filter labels.")
```

ArtieReus: I used to compare with `stringContaining` to avoid to fix the test if the text changes since there is jus one warning for the parseInitialFilters.

ArtieReus: example:
https://github.com/cloudoperators/juno/blob/2a037feaefa10e480fe599b5053037dd0ec7b453/apps/greenhouse/src/lib/plugin.test.js#L22
while the text is bit longer:
https://github.com/cloudoperators/juno/blob/2a037feaefa10e480fe599b5053037dd0ec7b453/apps/greenhouse/src/lib/plugin.js#L33

andypf (approved):
👍

ArtieReus (approved):
lgtm

PR #503, opened 2024-10-08
chore(ui): remove dependency to assets server (fonts)

andypf:
# Summary

This PR removes the dependency on the assets server. Previously, we loaded fonts from the assets server, but since it has been decommissioned, we had to integrate the fonts directly into the project, which has been done in this PR.
 
# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


barsukov (commented):
I will comment it from my side is good but @edda please have a look

packages/ui-components/src/components/StyleProvider/Fonts.js
```diff
@@ -11,39 +11,27 @@
...
-      const link1 = document.createElement("link")
-      link1.rel = "preconnect"
-      link1.href = "https://fonts.googleapis.com"
-
-      const link2 = document.createElement("link")
-      link2.rel = "preconnect"
-      link2.href = "https://fonts.gstatic.com"
-      link2.crossOrigin = "anonymous"
-
-      const link3 = document.createElement("link")
```
edda, line 32: I believe we need this link. Afaik in order for webfonts to work properly within the shadow root they need to be included via link in the head (which is what this `link3` does) and additionally inside of the shadow root they need to be included again inline (which is what happens in line 34). 

This might be outdated information, but please double check again.

edda (changes requested):

packages/ui-components/src/components/StyleProvider/Fonts.js
andypf: I'll check, thanks

andypf: Since we are now embedding the fonts as text, in my opinion, they no longer need to be loaded separately in the head. Currently, we either load the fonts inline in the shadow root as text or in the head.

edda (approved):

hgw77 (approved):

hodanoori (approved):

PR #510, opened 2024-10-10
fix(ui): fix broken pageheader layout (#506, #473)

### LINKED ISSUE #473: [Bug](juno-ui): PageHeader doesn't display children correctly anymore
**Description**

With the latest version, PageHeader seems to lost the ability to host children objects - this is a breaking change and I cannot find a replacement for it.

**Steps to Reproduce**

1. `use <PageHeader> with children, e.g. <PageHeader><h1>Hello World</h1></PageHeader>`

**Expected Behavior**
Children to embedded into the pageheader content

**Screenshots**

image

**Environment**

- OS: MacOS
- Browser: Chrome
- Version: 129.0.6668.71

**Additional Context**

### LINKED ISSUE #506: PageHeader broken in example app
**Description**
Currently, the `PageHeader` appears to be broken in the example app: There is too much space to the right of the logo effectively pushing the application title to the right; the options (user name, log-out button) are not visible at all.

**Steps to Reproduce**
Run the example app.

**Expected Behavior**
The page header should look as designed.

**Screenshots**
Image

**Environment**
Any OS, any browser

franzheidl:
# fix(ui): fix broken pageheader layout (#506, #473) – Summary
The pageheader layout was broken. When passing children (such as user name, log-out/in buttons) three child nodes were rendered into a grid container with only two templatce columns defined. Furthermore, we used to have overflow-hidden set on the header, thus the options became not only rendered in the wrong place, but also were invisible.

# Changes Made
* Make sure the `.juno-pageheader-inner` element only ever has two child nodes, that were explicitly named `juno-pageheader-logo-container` and `.juno-pageheader-content-container`. 
* Any children are now rendered into a `.juno-pageheader-options` container **inside** `.juno-pageheader-content-container`.
* remove hidden overflow from `.juno-pageheader-inner` to make options visible even if the layout is broken to increase our chance to spot any issues earlier, and keep the header at least functional.
* Always render the `.juno-pageheader-heading` container, even if no heading is supplied to make sure it is the first grid column in the `.juno-pageheader-inner` grid container
* Added stories with children/options so we can visually check in storybook

## Previous
```
<div class="juno-pageheader">
  <!-- Grid container: -->
  <div class="juno-pageheader-inner">
       <!-- First column: -->
      <div class="juno-pagheader-logo-container"><!-- logo goes here --></div>
      <!-- Second column: -->
      <div>App Title</div>
      <!-- When options were passed, we rendered a third column that effectively broke the two-column layout: -->
      <div><!-- options go here --></div>
    </div>
  </div>
</div>
```

## Fixed
```
<div class="juno-pageheader">
   <!-- Grid container: -->
  <div class="juno-pageheader-inner">
      <!-- First column: -->
      <div class="juno-pagheader-logo-container"><!-- logo goes here --></div>
      <!-- Second column: -->
      <div class="juno-pageheader-content">
         <div class="juno-pageheader-heading">App Title</div>
         <div class="juno-pageheader-options"><!-- options go here --></div>
      </div>
    </div>
  </div>
</div>
```

# Related Issues
fixes #506 , fixes #473

# Screenshots (if applicable)

Broken header layout
Screenshot 2024-10-09 at 11 13 46
 in example app with invisible options:

Fixed `PageHeader` layout:
Screenshot 2024-10-10 at 09 39 22

# Testing Instructions
Visually check in storybook and example app.

# Checklist

- [ ] **N/A:** I have added tests that prove my fix is effective or that my feature works – opted against introducing tests for where elements are rendered
- [ ] **N/A:** I have made corresponding changes to the documentation (if applicable)


andypf (dismissed):
LGTM

packages/ui-components/src/components/PageHeader/PageHeader.stories.tsx
```diff
@@ -64,24 +69,34 @@ export const WithHeading: Story = {
   },
 }
 
+export const WithHeadingAndOptions: Story = {
```
edda, line 72: Options is the internal name. I think it'd be better to say "children" here because that's how users have to use it
```suggestion
export const WithHeadingAndChildren: Story = {
```

edda (changes requested):
I feel like there are some potential edge cases with the heading/options divs in the content

packages/ui-components/src/components/PageHeader/PageHeader.component.tsx
```diff
@@ -35,13 +34,21 @@ const logoContainerStyles = `
...
-const headingStyles = (clickable: boolean) => {
-  return `
-    jn-text-lg
-    jn-text-theme-high
-    ${clickable && "jn-cursor-pointer"}
-    `
-}
+const contentContainerStyles = `
+  jn-grid
+  jn-grid-cols-[1fr,minmax(0,max-content)]
```
edda, line 39: If no heading exists but there are children then the children will be dumped into the first grid column here because the div containing the heading is conditionally rendered (i.e. it doesn't exist at all if no heading was passed). 

I think we might be able to fix this by defining named grid areas and ensuring the children always get rendered in the second column. Or, probably easier, always render the heading div, even if it's empty.

edda (changes requested):

barsukov (commented):
I could not say anything specific on this one :) LGTM from my side but will leave it as comment maybe we could cover this case in tests as well.

packages/ui-components/src/components/PageHeader/PageHeader.stories.tsx
franzheidl: Agreed.

packages/ui-components/src/components/PageHeader/PageHeader.component.tsx
franzheidl: Absolutely! 

I have removed the conditional rendering logic for the `.juno-pageheader-heading` div, so it will always be there, preventing the options from slipping into the second grid column and thus preserving their position ont he right of the header at all times.

edda (approved):

barsukov (commented):
From my side no issues so far, but will wait for other reviewers

barsukov (approved):

PR #511, opened 2024-10-10
chore(ui): migrate Message component to Typescript

guoda-puidokaite:
# Summary

- Migrated the `Message` component to `TypeScript`, enhancing code quality, maintainability, and scalability.
- Refactored the component for cleaner, more maintainable code.

# Changes Made

**TypeScript Migration**:

- Converted `Message` component files to `TypeScript`.

**Bug Fix:**

- Auto-dismiss is now correctly triggered only for values greater than or equal to one.

**Refactoring**:

**Testing**:

- Added tests for bug fix.
- Fixed tests that weren't checking for the close button in case of duplicates.
- Organised tests by what they check for better readability and maintenance.

**Cleanup:**

**Documentation:**

Screenshot 2024-10-14 at 08 13 31

**Release:**

# Related Issues

- https://github.com/cloudoperators/juno/issues/238

# Testing Instructions

1. Run `Storybook`
2. Run `Message` tests

# Checklist


guoda-puidokaite (commented):
Bug found: A timeout of zero or a negative number initiates a message dismissal. [Mentioned in main issue](https://github.com/cloudoperators/juno/issues/238#issuecomment-2413302523).

packages/ui-components/src/components/Message/Message.types.ts
```diff
@@ -0,0 +1,27 @@
...
+  /** Optional variant for message appearance e.g. 'success'. Default is 'info'. */
+  variant?: MessageType
+  /** Optional: If true, the message will have a 'close' button. Default is false. */
+  dismissible?: boolean
+  /** Optional: If true, the message will auto-dismiss after 10 seconds or the specified autoDismissTimeout. */
+  autoDismiss?: boolean
+  /** Optional timeout in milliseconds after which the message auto-dismisses. Default is 10000 (10 seconds). */
+  autoDismissTimeout?: number
+  /** Optional handler function called once the message is dismissed */
+  onDismiss?: () => void
```
andypf, line 22: This results in the type being displayed in a user-unfriendly way in Storybook.

image

I’m not an expert in TypeScript yet, but could `React.MouseEventHandler` be useful in this case?

andypf (commented):
Good Job Guoda 👍. However, I found a couple of issues.

andypf:
> Bug found: A timeout of zero or a negative number initiates a message dismissal

Did you fix it?

packages/ui-components/src/components/Message/Message.component.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useEffect, useState, useRef } from "react"
+
+import { Icon } from "../Icon"
+
+import { MessageProps } from "./Message.types"
+import { getMuiIcon } from "./utils/iconUtils"
```
edda, line 11: I don't see a big advantage of moving the icon logic into utils if only this one component makes use of it. However with small adjustments it might be reusable for all components that use icons. So instead of checking `MessageType` make a more generic type like `SemanticVariant` or something, then move the iconUtils up a few levels to `src/utils`. 

Though my suggestion would be to not do the generalization now but rather do this in a new PR and then use this new shared util in all components with icons. Though it might be more useful to do this after everything (or at least all components that use icons) have moved to Typescript :)

packages/ui-components/src/components/Message/Message.component.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { initiateAutoDismiss, clearAutoDismissTimeout } from "./utils/timeoutUtils"
```
edda, line 12: I would prefer to keep this inline in the `Message` component right now. Mostly to be in line with other similar components where it is like this, too. It makes it harder to see what the functions do because you have to jump to another file. 

In general I'm not against this type of refactoring but then we need to look at it holistically and do the same thing in all components where it applies to ensure consistency.

edda (changes requested):

packages/ui-components/src/components/Message/Message.types.ts
edda: Check e.g. the `Button` component. It has been moved to TS and also has click event handlers

guoda-puidokaite: Thank You @andypf and @edda.
I think that `React.MouseEventHandler` would be incorrect in this context. 
In the `Button` component, it's tied to the specific mouse event `onClick`, so `React.MouseEventHandler<EventTarget>` is correct.
But here, the `onDismiss` function is a custom handler. It should just be a function that takes no arguments and returns nothing - `(() => void)`.

packages/ui-components/src/components/Message/Message.component.tsx
guoda-puidokaite: Thank You for the review, Esther. No problem, I can do that. For me, it makes the component code very hard to read. I like to open a component file and see quickly what it does. Other helpers, especially css variables etc. are secondary to that, in my opinion. 

Overall it was a really nice way for me to play around with the code and get familiar with it. 🤪

But the most important thing indeed is consistency. If the team prefers to structure code this way across all components and find it easier to understand the full logic in one place, I agree we should keep it this way. Thank You. 🚀

packages/ui-components/src/components/Message/Message.component.tsx
guoda-puidokaite: I didn't want to touch the other components and just try it out here. 🙈 But I'll revert all externalisation and we can discuss it in the future. Thank You.

packages/ui-components/src/components/Message/Message.component.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {
```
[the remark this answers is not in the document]
guoda-puidokaite: Will do. I'd at the very least consider placing the main component at the top, to make it easier to quickly understand the core logic of the file without scrolling past utility functions and type definitions. 🙈 And due to beauty of hoisting still be good. But again convention, so we can leave it as is.

packages/ui-components/src/components/Message/Message.types.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export type MessageType = "info" | "warning" | "danger" | "error" | "success"
```
guoda-puidokaite, line 6: @edda Should we move types into the main component? Other components do so.

guoda-puidokaite:
Hey, @andypf and @edda. I made the changes and added comments for everything else. Thank You for the review. 🚀

Changes made:
- Bug fixed: Auto dismiss is triggered for values less than one
- Added tests for fix and grouped tests by type
- Updated the `changeset` message to include bug fix
- Moved all externalised `css`, `icon` and `timeout` logic back to the main component

packages/ui-components/src/components/Message/Message.types.ts
andypf: Got it, let’s wait until Wowa is here, maybe he has an idea.

andypf (dismissed):
LGTM

packages/ui-components/src/components/Message/Message.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  // VARIANTS
+
+  test("renders an info Message by default if no variant is provided", () => {
+    render(<Message data-testid="my-message" />)
+    expect(screen.getByTestId("my-message")).toHaveClass("juno-message-info")
+    expect(screen.getByRole("img")).toHaveClass("jn-text-theme-info")
+  })
+
+  test.each`
```
barsukov, line 66: i would like to discuss this point it is very similar to a Badge PR :)

packages/ui-components/src/components/Message/Message.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  test("renders a Message without a dismiss button by default", () => {
+    render(<Message data-testid="my-message" />)
+    expect(screen.queryByRole("button", { name: /close/i })).not.toBeInTheDocument()
+  })
+
+  test("renders a dismissible Message when 'dismissible' is true", async () => {
+    render(<Message data-testid="my-message" dismissible={true} />)
+
+    const user = userEvent.setup()
```
barsukov, line 98: Do we need a setup here?

packages/ui-components/src/components/Message/Message.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+      expect(mockOnDismiss).toHaveBeenCalledTimes(1)
+    })
+  })
+
+  test("does not auto-dismiss when 'autoDismissTimeout' is set to zero", async () => {
+    render(<Message data-testid="my-message" autoDismiss={true} autoDismissTimeout={0} />)
+    expect(screen.getByTestId("my-message")).toBeInTheDocument()
+
+    // Wait to check that the message did not auto-dismiss
+    await new Promise((resolve) => setTimeout(resolve, 1000))
```
barsukov, line 150: Do we a promise here? is not waitFor or act are made for?

packages/ui-components/src/components/Message/Message.component.tsx
```diff
@@ -0,0 +1,210 @@
...
+ */
+
+import React, { useEffect, useState, useRef } from "react"
+
+import { Icon } from "../Icon"
+import { KnownIcons, KnownIconsEnum } from "../Icon/Icon.component"
+
+import { MessageProps, VariantType } from "./Message.types"
+
+// CSS UTILS
```
barsukov, line 13: Lets discuss this component on a call.

barsukov (changes requested):
Good job! I have few comments left, lets jump on a call and discuss :) 🔢

packages/ui-components/src/components/Message/Message.test.tsx
guoda-puidokaite: Hmmm, I don't think so. Added due to this error (`waitFor` was there without `await`): 
Screenshot 2024-10-14 at 14 00 57
To my understanding, `waitFor` repeatedly calls what's inside until the condition is met or the timeout is reached. 
But it's still a `Promise`, so we need to stop execution in our code for it to be fulfilled.
But good point, now that I reflect on it... Can't we just do below? Much more readable.
```
await user.click(screen.getByRole("button", { name: /close/i }))
expect(screen.queryByTestId("my-message")).not.toBeInTheDocument()
```
However, the first way is used all over our code base...

packages/ui-components/src/components/Message/Message.test.tsx
guoda-puidokaite: We probably don't need it for such a simple test, though it's recommended for more reliable and accurate test results. It helps handle asynchronous state updates and side effects. But just fyi: We use it across our code base, I didn't add it here. 🤪

packages/ui-components/src/components/Message/Message.types.ts
guoda-puidokaite: Done

packages/ui-components/src/components/Message/Message.component.tsx
```diff
@@ -0,0 +1,259 @@
...
+  warningBg: `jn-bg-theme-message-warning`,
+
+  danger: `jn-border-theme-message-danger`,
+  dangerBg: `jn-bg-theme-message-danger`,
+
+  success: `jn-border-theme-message-success`,
+  successBg: `jn-bg-theme-message-success`,
+}
+
+export type VariantType = "info" | "warning" | "danger" | "error" | "success"
```
barsukov, line 64: I would like to rename it here same way as we did with badge

guoda-puidokaite: Ah, good catch!

packages/ui-components/src/components/Message/Message.component.tsx
```diff
@@ -0,0 +1,259 @@
...
+      return messageVariantStyles.danger
+    default:
+      return messageVariantStyles.default
+  }
+}
+
+// Determine if a given string corresponds to a known icon
+const isValidIcon = (icon: string): icon is KnownIcons => {
+  const validIconNames: Set<KnownIcons> = new Set(Object.values(KnownIconsEnum))
+  return validIconNames.has(icon as KnownIcons)
```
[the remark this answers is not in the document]
guoda-puidokaite: Yes, just because it's a `string`. But because we know the result will be valid `KnownIcon`, should be fine to cast.

barsukov (approved):
Great! 🥇

andypf (approved):
👍

edda (approved):
🚀

PR #519, opened 2024-10-14
feat(heureka): extends test

hodanoori:
# Summary

Add a unit test for global slice store.

# Related Issues

- [ ] fixes https://github.com/cloudoperators/heureka/issues/287

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


andypf (dismissed):
LGTM

ArtieReus (commented):
Good job!! I like tests :)
I put an improvement but not request. I think testing slices decoupled from the store not a good pattern since we build dependencies between slices. What do you think?

apps/heureka/src/lib/slices/createGlobalsSlice.test.js
```diff
@@ -0,0 +1,116 @@
...
+import { create } from "zustand"
+import createGlobalsSlice from "./createGlobalsSlice"
+import constants from "../../components/shared/constants"
+
+describe("createGlobalsSlice", () => {
+  let useStore
+  let initialState
+
+  beforeEach(() => {
+    useStore = create((set, get) => ({
```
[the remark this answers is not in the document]
ArtieReus: You can even create your store with options as I do also here:
https://github.com/cloudoperators/juno/blob/main/apps/supernova/src/lib/createFiltersSlice.test.js#L48

hodanoori:
> Good job!! I like tests :) I put an improvement but not request. I think testing slices decoupled from the store not a good pattern since we build dependencies between slices. What do you think?

Good suggestion, I re-write the test to run within the actual store context. Please have a look again.

ArtieReus (dismissed):
Now it looks very good!

apps/heureka/src/App.test.jsx
```diff
@@ -4,17 +4,21 @@
...
 // https://reactjsexample.com/an-extension-of-dom-testing-library-to-provide-hooks-into-the-shadow-dom/
 import { screen } from "shadow-dom-testing-library"
 import App from "./App"
 
 describe("App", () => {
   test("should render the App component", async () => {
-    render(<App id="123" />)
-    const loginTitle = screen.queryAllByShadowText(/HEUREKA/i)
-    expect(loginTitle.length > 0).toBe(true)
+    await act(async () => {
```
barsukov, line 15: Why you do here a render under the act it is usually it is only used on state change or any side effects(like clicks, etc)

apps/heureka/src/lib/slices/createGlobalsSlice.test.jsx
```diff
@@ -0,0 +1,275 @@
...
+  it("should update showPanel", () => {
+    const { result } = renderHook(
+      () => ({
+        actions: useGlobalsActions(),
+        showPanel: useGlobalsShowPanel(),
+      }),
+      { wrapper }
+    )
+
+    act(() => {
```
barsukov, line 155: I am not sure that we need an act here ...

barsukov: I mean everywhere where you use it in a whole test file

barsukov (changes requested):
Good start 👍 We need a little improvements here:
1) Use act only when it is neccessary
2) Maybe grouped some tests in Slice test file

apps/heureka/src/App.test.jsx
hodanoori: To solve the following warning, I used act. If not, what is your suggestion?
image

apps/heureka/src/lib/slices/createGlobalsSlice.test.jsx
hodanoori: The same here, I have to use act, otherwise when I use the store actions directly, I get the following error, because here I render StoreProvider to instatiate the store context. 
` Warning: An update to TestComponent inside a test was not wrapped in act(...).
│ 
│ When testing, code that causes React state updates should be wrapped into act(...):
│ 
│ act(() => {
│   /* fire events that update state */
│ });
│ /* assert on the output */
│ 
│ This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
│     at TestComponent (/workspace/juno/node_modules/@testing-library/react/dist/pure.js:307:5)
│     at StoreProvider (/workspace/juno/apps/heureka/src/components/StoreProvider.jsx:19:26)
│     at wrapper (/workspace/juno/apps/heureka/src/lib/slices/createGlobalsSlice.test.jsx:14:22)`

apps/heureka/src/components/StoreProvider.test.jsx
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { act } from "react"
+import { renderHook } from "@testing-library/react"
+import {
+  StoreProvider,
+  useGlobalsEmbedded,
+  // useGlobalsApiEndpoint,
```
andypf, line 11: please remove this comments

andypf (approved):

andypf (approved):
👍

barsukov (approved):
LGTM 👍

PR #523, opened 2024-10-15
feat(docs): adds a brd for carbon app loading mechanism

andypf:
# Summary

This PR adds a Business Requirements Document (BRD) that outlines the requirements for the Carbon Architecture concerning the hosting and loading of extensions. The document details key features, including:

- Extendability: Enables seamless integration of third-party apps, whether colocated or remotely hosted.
- Scalability: Supports simultaneous loading and management of multiple extensions without impacting performance.
- Flexibility: Utilizes a manifest-driven configuration approach for managing various types of extensions.
- Isolation: Ensures that extensions operate independently, maintaining security and functional boundaries.

The document covers important aspects such as routing mechanisms, loading processes, communication protocols, and security considerations. This foundational architecture will guide the implementation of a robust extension framework within the Carbon platform, paving the way for future development and integration.


docs/proposals/004_carbon_brd.md
```diff
@@ -0,0 +1,109 @@
...
+Security is a critical aspect of loading external extensions in **Carbon**, particularly when dealing with remote or iframe-based apps.
+
+#### Requirements:
+
+- **Sandboxing for Iframes:** All iframe-based extensions should be sandboxed, enforcing restrictions like `allow-scripts`, `allow-same-origin`, and other relevant flags to limit potential security risks (e.g., cross-site scripting, data leakage).
+- **CORS and Content Security Policy (CSP):** For remote extensions, the platform must enforce proper CORS headers and CSP to ensure that only authorized domains are allowed to load and interact with the platform.
+
+- **Authentication and Authorization:** Extensions must adhere to the **Carbon** platform's authentication and authorization mechanisms, ensuring that only authenticated users with appropriate permissions can load or interact with specific extensions.
+
+- **Dependency Security:** The manifest file or configuration source must define any external dependencies required by the extension. These dependencies should be validated for security vulnerabilities before being loaded into the platform.
```
taymoor89, line 103: validating or checking security vulnerabilities of external dependencies might not be possible for **remote extensions** because if I understood correctly they would already come as final executable/importable bundle meaning all of the external dependencies would already be packaged.

docs/proposals/004_carbon_brd.md
```diff
@@ -0,0 +1,109 @@
...
+
+### 1. Routing
+
+Routing is essential for isolating and providing direct access to each extension within the **Carbon** architecture.
+
+#### Requirements:
+
+- **Dedicated Route per App:** Each extension should have its own unique route, such as `/app/:appId`, which should support both local and remote navigation.
+- **Isolated Routing:** Extensions should not modify or interfere with the base URL of the host app. Any internal routing within an extension must remain scoped to that extension, ensuring that changes in URL or navigation do not affect the host's routing logic.
+
```
edda, line 37:
```suggestion
- **Cross Linking Support:** While extensions should not be able to modify the base URL directly the host app should provide an interface with which extension can cross link to each other
```

docs/proposals/004_carbon_brd.md
```diff
@@ -0,0 +1,109 @@
...
+- **Support for Route Parameters:** Routes should support parameters, allowing apps to receive specific data or inputs (e.g., `/app/:id/:tab`). These parameters should be passed to the app as props or through another mechanism (e.g., context or query parameters).
+
```
edda, line 39:
```suggestion
- **Support for query parameters:** Routes must support query parameters, e.g. `/app?foo=blah` or `/app/:id?foo=blah`
```

docs/proposals/004_carbon_brd.md
```diff
@@ -0,0 +1,109 @@
...
+- **Deep Linking Support:** The **Carbon** system must support deep linking, allowing users to navigate directly to a specific state or view within an extension using a complete URL path (e.g., `/app/:appId/view/:viewId`).
+
```
edda, line 41:
```suggestion
- **App URL State Storage:** The previous URL state of an app should be saved (e.g. in browser local storage) when navigating to a different app and retrieved when returning to it. This behaviour should be configurable (turn on/off) per app. On initial load of the host app all previously stored app states should be invalidated.
```

docs/proposals/004_carbon_brd.md
```diff
@@ -0,0 +1,109 @@
...
+2. **Remote Extensions:**
+   - **Definition:** Extensions hosted externally, often served from a CDN or remote server.
+   - **Loading via URL:** Remote extensions should be dynamically loaded based on the URL provided in the manifest, allowing for flexible integration of micro frontends or remotely hosted modules.
+   - **Iframes as a Loading Mechanism:** If an extension is loaded via an iframe, it must be securely sandboxed, and communication should be facilitated through a defined API (see Communication section below).
+
+**Standardized Mounting Interface**
+
+- **Mount API:** The **Carbon** platform must provide a consistent mounting interface for all extension types (e.g., `mount(container, props)`). This interface should allow colocated and remote extensions to be mounted dynamically within a specific DOM container while passing the necessary props and configuration details from the manifest.
+- **Lazy Loading:** The **Carbon** system should support lazy loading of extensions, allowing resources to be fetched and initialized only when needed, optimizing performance and reducing the initial load time.
+
```
edda, line 76:
```suggestion
**Isolation and Error Boundaries**
Extensions must be loaded into an isolated container, either ShadowRoot (preferred) or iframe. Isolation of CSS styles between host app and extension must be ensured. Apps must be wrapped in an ErrorBoundary to ensure runtime errors of the extensions/apps don't cause problems for the remaining extensions or host app
```

docs/proposals/004_carbon_brd.md
ArtieReus: > validating or checking security vulnerabilities of external dependencies might not be possible for **remote extensions** because if I understood correctly they would already come as final executable/importable bundles meaning all of the external dependencies would already be packaged inside those bundles.

I believe the intention is that no extensions will be loaded or rendered until the user is authenticated in the shell app. Once authenticated, any necessary properties can then be passed to the extensions as needed.

docs/proposals/004_carbon_brd.md
ArtieReus: > We probably want something like this, too.

We decided to include these details in the Interface section, but I believe it's worth mentioning them here as well.

docs/proposals/004_carbon_brd.md
ArtieReus: I agree also with @edda. I thought we mentioned also yesterda :)

docs/proposals/004_carbon_brd.md
andypf: I appreciate the discussion here; it’s exactly what this space is meant for! 👍 My original thought was that bundling the extensions at build time would allow us to run checks on the interface and code quality using tools like Trivy. However, I understand that this approach wouldn’t be feasible for iframe-based extensions. For those, we would likely need to rely on a manual process for validation.

taymoor89: exactly for third party extension running in an iframe or coming as importable bundles we can't verify their dependencies so it is up to the user to do the due diligence before integrating them to the system. From our side we can only render them in a sandboxed/secure environment so they don't mess up with the shell or other loaded extensions.

barsukov (approved):

taymoor89 (approved):

PR #545, opened 2024-10-22
chore(ui): migrate Breadcrumb components to Typescript

guoda-puidokaite:
# Summary

- Migrated `Breadcrumb` and `BreadcrumbItem` components to `TypeScript`, enhancing code quality, maintainability, and scalability.
- Enhanced unit tests, covering more edge cases.
- Improved documentation with clearer descriptions of the component and its props.

# Changes Made

**TypeScript Migration**:

- Converted `Breadcrumb` and `BreadcrumbItem` component files to `TypeScript`.

**Refactoring**:

**Testing**:

- Updated `BreadcrumbItem` test case `renders an icon as passed` (**please double check if correct, was failing**)
- Added more Edge Cases:
  - `Breadcrumb` Component
    - renders children elements correctly
    - renders separators between breadcrumb items
  - `Breadcrumb Item` Component
    - renders without any props
    - does not render a link if 'href' is null
    - adds all passed props to the root element
    - renders correctly with both active and disabled states
    - renders as a link and executes onClick handler
    - renders only children if both children and icon are passed
    - renders custom className without overriding component-specific classes
- Organised tests by functionality for better readability and easier maintenance.

**Cleanup:**

- Update `ui-components/src/index` to point to .`ts` components for all components converted to `.ts` so far.

**Documentation:**

Screenshot 2024-10-22 at 14 45 17
Screenshot 2024-10-22 at 14 48 39

**Release:**

# Related Issues

- https://github.com/cloudoperators/juno/issues/215

# Testing Instructions

1. Run `Storybook`
2. Run `Breadcrumb` and `BreadcrumbItem` tests

# Checklist


guoda-puidokaite:
Closes https://github.com/cloudoperators/juno/issues/215

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+  })
+
+  describe("Handling Props", () => {
+    test("renders a breadcrumb item with text as passed", () => {
+      render(<BreadcrumbItem data-testid="breadcrumbitem" />)
+      expect(screen.getByTestId("breadcrumbitem")).toBeInTheDocument()
+      expect(screen.getByTestId("breadcrumbitem")).toHaveClass("juno-breadcrumb-item")
+    })
+
+    test("renders an icon as passed", () => {
```
guoda-puidokaite, line 36: Test was failing. I think the structure changed in `Icon` during migration to `Typescript`. 
Please double check that how I updated the test is what's expected. 
Otherwise, there might be a bug in `Icon`.

guoda-puidokaite: This test probably belongs in `Icon` and not `BreadcrumbItem` too...

barsukov: what exactly failing here?

barsukov (commented):
One comment regarding the test should we solve together and I think we could move on with this one. Good job 👍

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
guoda-puidokaite: CC: @edda

barsukov (dismissed):
As we spoke before I found it good for me. Please ask the second 👀

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
edda: hmm, regarding whether this test should be here: yes, it should. It's not about testing that the image has the correct title attribute, that is just the way that was chosen here to verify that not only is an icon rendered, it's also the correct one. You're right, it's more of an integration test but if it succeeds then it verifies also, that nothing weird is happening within the BreadcrumbItem component.

Regarding the failing test, this is unexpected. The Icon component hasn't changed structurally with the migration. My guess is that the assertion `toHaveAttribute` doesn't work the same in vitest as it did in jest. The way `Icon` works is that we build a component for each svg and add a bunch of props to the component. One of the props is `title`. However the rendered svg doesn't get a title attribute, instead a `<title>` tag is added inside of the svg. In jest the `toHaveAttribute("title", ..)` assertion seems to have correctly evaluated the title tag inside the svg and I'm guessing in vitest it doesn't. Could you try to verify that that is the problem? In that case we could instead of the title assert on the `alt` attribute.

edda (changes requested):

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
guoda-puidokaite: Indeed, good point that it could be jest vs. vite. I believe so. In the updated test, I'm querying for the `title` element inside the image/svg (instead of attribute), is that not enough?🙈

edda: My main concern regarding this is that it assumes the icon to be an svg. If that ever changes then this test would break. That's why I suggested asserting on the `alt` attribute instead since that will be present irrespective of image type.

guoda-puidokaite: Ah, got it. Maybe the naming is the problem. It's just searching for an `img`, I think? Not necessarily an svg. But I update the test to use `alt`, more straight forward.

barsukov (approved):
LGTM :)

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
edda: No, it searches for an element that has the aria role `img`. This encompasses everything that's an image, irrespective of type.

PR #547, opened 2024-10-22
chore(ui): migrate AppShellProvider and CodeBlock to TypeScript

gjaskiewicz:
# Summary

Migrate AppShellProvider and CodeBlock to TypeScript

# Changes Made

- migrate AppShellProvider to TypeScript
- migrate CodeBlock to TypeScript
- expose deprecated_js version of AppShellProvider in index.js - temporary fix until this component is removed
- add replacement for CodeBlock in stories (to be replaced after migration of this component)

# Related Issues

- Issue 1: [[link to issue](https://github.com/cloudoperators/juno/issues/212)]
- Issue 2: [[link to issue](https://github.com/cloudoperators/juno/issues/219)]

# Testing Instructions

1. `npm i`
2. `npm run storybook`

# Checklist


packages/ui-components/src/index.js
```diff
@@ -7,7 +7,7 @@ import "./global.scss"
 export { AppBody } from "./components/AppBody/index.js"
 export { AppIntro } from "./components/AppIntro/index.js"
 export { AppShell } from "./components/AppShell/index.js"
-export { AppShellProvider } from "./components/AppShellProvider/index.js"
+export { AppShellProvider } from "./deprecated_js//AppShellProvider/index.js"
```
andypf, line 10: double slash

packages/ui-components/src/components/AppShellProvider/AppShellProvider.component.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * This provider acts as a wrapper for Juno apps. It renders a StyleProvider and PortalProvider
+ */
+export const AppShellProvider = ({
+  shadowRoot = true,
+  shadowRootMode = "open",
+  stylesWrapper = "inline",
+  theme = null,
+  children,
+}: AppShellProviderProps) => {
+  const Wrapper = React.useCallback(
```
andypf, line 22: We now have the opportunity to refactor it properly by extracting it into a regular component, outside of this one, without the need for useCallback.

packages/ui-components/src/components/AppShellProvider/AppShellProvider.stories.tsx
```diff
@@ -4,9 +4,16 @@
  */
 
 import React from "react"
-import { AppShellProvider } from "."
-import { CodeBlock } from "../CodeBlock/index.js"
+// import { CodeBlock } from "../CodeBlock/index.js"
```
andypf, line 7: can we remove this line?

barsukov (changes requested):
Please fix the Andreas comments and also from my side change index.js to use a deprecated one. 

When we migrate Combobox and also Select we will delete all deprecated related files and change the exports in the index.js

barsukov:
We close caause we need first to migrate the jsonviewer component and also codeblock

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -149,11 +154,11 @@ export const CodeBlock = ({
         ""
       )}
       {lang === "json" ? (
-        <JsonViewer data={content} expanded={3} theme={jsonTheme} style={jsonViewStyles} />
+        <JsonViewer data={content as object} expanded={3} theme={jsonTheme} style={jsonViewStyles} />
```
barsukov, line 157: Could we get rid of this casting here?

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -149,11 +154,11 @@ export const CodeBlock = ({
...
       ) : (
         <pre className={`juno-code-block-pre ${preStyles(wrap)} ${sizeStyles(size)}`}>
           <code className={`${codeStyles}`} ref={theCode}>
-            {content || children}
+            {(content || children) as React.ReactNode}
```
barsukov, line 161: Could we get rid of this casting here?

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -170,21 +175,23 @@ export const CodeBlock = ({
...
-CodeBlock.propTypes = {
+type CodeBlockSize = "auto" | "small" | "medium" | "large"
+
+export interface CodeBlockProps {
   /** The content to render. Will override children if passed. */
-  content: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+  content?: string | object
   /** The children to render. Will be overridden by content prop if passed as well.  */
-  children: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
+  children?: React.ReactNode | object
```
barsukov, line 184: do we need it here as an object?

barsukov (changes requested):
few comments

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
gjaskiewicz: Done

barsukov (approved):
LGTM

TilmanHaupt (approved):

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -117,21 +116,27 @@ export const CodeBlock = ({
...
-  const theCode = useRef(null)
+  const theCode = useRef<HTMLElement | null>(null)
 
   const handleCopyClick = () => {
-    const textToCopy = lang === "json" ? JSON.stringify(content || children) : theCode.current.textContent
-    navigator.clipboard.writeText(textToCopy)
+    const textToCopy = lang === "json" ? JSON.stringify(content || children) : theCode.current!.textContent
+    if (textToCopy) {
+      navigator.clipboard.writeText(textToCopy).catch(() => {
+        console.warn("Cannot copy text to clipboard")
```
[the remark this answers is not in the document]
guoda-puidokaite: Perfect. Don't worry about it in this PR, @gjaskiewicz. Adding this to an improvement EPIC instead.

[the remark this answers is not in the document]
guoda-puidokaite: Created placeholder issue: https://github.com/cloudoperators/juno/issues/572

packages/ui-components/src/components/CodeBlock/CodeBlock.test.tsx
```diff
@@ -8,28 +8,28 @@ import { render, screen } from "@testing-library/react"
 import { CodeBlock } from "./index"
 
 describe("CodeBlock", () => {
```
guoda-puidokaite, line 10: Would be great during these refactors to improve testing. Some missing edge cases could be:
- Component rendering when both `content` and `children` are `null` or `undefined`.
- Rendering (fallback and default) behaviour for different `lang` prop values, including unsupported languages.
- Clipboard copy functionality to ensure correct text is copied and the "Copied!" message appears.
- Ensuring `timeoutRef` is properly cleared when the component unmounts to prevent memory leaks.
- Ensuring `theCode.current` is correctly assigned and not accessed when `null`.

guoda-puidokaite (commented):
Would be great to improve the tests of both components. Especially `AppShellProvider`, it only has a few. If you've the time, of course. 🙈

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -4,23 +4,22 @@
  */
 
 import React, { useState, useRef } from "react"
-import PropTypes from "prop-types"
-import { JsonViewer } from "../../deprecated_js/JsonViewer/JsonViewer.component"
-import { Icon } from "../../deprecated_js/Icon/index"
+import { JsonViewer } from "../JsonViewer/JsonViewer.component"
+import { Icon } from "../Icon/index"
```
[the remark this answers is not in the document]
guoda-puidokaite: https://github.com/cloudoperators/juno/pull/571#issuecomment-2443769757

guoda-puidokaite (changes requested):
Thank You for the changes!!! 🚀 Last suggested improvements.

packages/ui-components/src/components/AppShellProvider/AppShellProvider.component.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { ShadowRoot } from "../ShadowRoot"
+import { PortalProvider } from "../PortalProvider"
+
+const Wrapper = ({ children, shadowRoot, shadowRootMode }: WrapperProps) => {
+  return shadowRoot ? (
+    <ShadowRoot mode={shadowRootMode}>
+      <>{children}</>
+    </ShadowRoot>
+  ) : (
+    children
```
guoda-puidokaite, line 18: I was wondering here if `children` need to be wrapped in a valid JSX tag https://github.com/cloudoperators/juno/pull/547#discussion_r1820773192

guoda-puidokaite (approved):
Looks good. @barsukov, please double check that recent suggested changes won't cause any issues in your eyes too. 💪

barsukov (approved):
LGTM when @edda is also ok then we could merge it I guess

gjaskiewicz:
@edda Can you take a look?

guoda-puidokaite (changes requested):
**(Requested Change)**

The `AppShellProvider` has a few breaking changes 🐞:

See screenshots (JS code on left, TS on right)

Screenshot 2024-10-31 at 11 11 45
Screenshot 2024-10-31 at 11 12 07

- Setting `theme-dark` and `theme-light` is broken.  
- `theme` Story **control** has changed from `string` to `object`.
- `theme` **type** has changed from `string` to `string|null`. If the `theme` can only be `theme-light` and `theme-dark`, let's create a **custom type** containing these (instead of just `string`)?
- `theme` **default value** should be `theme-dark`?

gjaskiewicz:
> **(Requested Change)**
> 
> The `AppShellProvider` has a few breaking changes 🐞:
> 
> See screenshots (JS code on left, TS on right)
> 
> Screenshot 2024-10-31 at 11 11 45 Screenshot 2024-10-31 at 11 12 07
> * Setting `theme-dark` and `theme-light` is broken.
> * `theme` Story **control** has changed from `string` to `object`.
> * `theme` **type** has changed from `string` to `string|null`. If the `theme` can only be `theme-light` and `theme-dark`, let's create a **custom type** containing these (instead of just `string`)?
> * `theme` **default value** should be `theme-dark`?

It seems like a potential breaking change in apps to me. If type discrepency is problem in storybook I could just change a control type there. Otherwise it works when theme-light or theme-dark is entered with quotes.

guoda-puidokaite:
@gjaskiewicz I think that the storybook documentation should stay unchanged after the migration. At least to me, it's confusing why I can't just enter text and would have to use quotes, since type is only `string|null`. Please also consider the other suggestions regarding custom type and type default, I think it would be an appropriate time for the improvements.

packages/ui-components/src/components/AppShellProvider/AppShellProvider.component.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * This provider acts as a wrapper for Juno apps. It renders a StyleProvider and PortalProvider
+ */
+export const AppShellProvider = ({
+  shadowRoot = true,
+  shadowRootMode = "open",
+  stylesWrapper = "inline",
+  theme = null,
```
edda, line 29:
```suggestion
  theme = "theme-dark",
```

packages/ui-components/src/components/AppShellProvider/AppShellProvider.component.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export interface AppShellProviderProps {
+  /** Whether the app is rendered inside a ShadowRoot. Only choose false if the app is meant to run as a stand-alone application. */
+  shadowRoot?: boolean
+  /** Shadow root mode */
+  shadowRootMode?: ShadowRootMode
+  /** Where app stylesheets are imported. This is only relevant if shadowRoot === false. If you use a ShadowRoot the styles must be inline. */
+  stylesWrapper?: AppShellStyleWrapper
+  /** theme: theme-dark or theme-light */
+  theme?: string | null
```
edda, line 59: We should make explicit in the type what the allowed values are, so rather than `string | null` let's to `"theme-light" | "theme-dark"` and then instead of defaulting to `null`, default to `"theme-dark"`. 
```suggestion
  theme?: "theme-dark" | "theme-light"
```

edda: Additionally this type as is breaks the control in storybook (as reported by @guoda-puidokaite )

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -117,21 +116,27 @@ export const CodeBlock = ({
   lang = "",
   className = "",
   ...props
-}) => {
+}: CodeBlockProps) => {
   const [isCopied, setIsCopied] = useState(false)
-  const timeoutRef = React.useRef(null)
+  const timeoutRef = React.useRef<NodeJS.Timeout | null>(null)
```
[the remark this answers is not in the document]
edda: Please fix as suggested by @guoda-puidokaite

packages/ui-components/src/components/StyleProvider/StyleProvider.component.tsx
```diff
@@ -149,7 +149,7 @@ export interface StyleProviderProps {
   /** What element to render as a wrapper, respectively where to render the StyleProvider.  */
   stylesWrapper?: StyleProviderStylesWrapper
   /** The name of the theme to render. */
-  theme?: string
+  theme?: string | null
```
edda, line 152: See comment above. Change to
```suggestion
  theme?: "theme-dark" | "theme-light"
```
Though ideally, since we use this in two places we define the type once and use it in both places.

edda (dismissed):

gjaskiewicz:
> @gjaskiewicz Nice that it's just a Storybook issue. I think that the storybook documentation should stay unchanged though after the migration. At least to me, it's confusing why I can't just enter text and would have to use quotes, since `type` is only `string|null`. Please also consider the other suggestions regarding **custom type** and **type default**, I think it would be an appropriate time for the improvements. 🚀

We could go with 3 options here:
- set it to enumerated type, if Esther confirms that setting arbitrary string should be no longer supported,
- allow arbitrary string, but disallow null, we could set default value to "theme-dark" then
- leave type as is, but change control in strorybook to string control rather than object control

As some of the options will cause change of behavior of the component it would be good if also @barsukov and @edda express their opinion

packages/ui-components/src/components/StyleProvider/StyleProvider.component.tsx
guoda-puidokaite: Indeed, I'd suggest a **custom type**, as [above](https://github.com/cloudoperators/juno/pull/547#issuecomment-2449722013) .

gjaskiewicz: As discussed f2f - I changed the type

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -170,21 +179,23 @@ export const CodeBlock = ({
...
-CodeBlock.propTypes = {
+type CodeBlockSize = "auto" | "small" | "medium" | "large"
+
+export interface CodeBlockProps {
   /** The content to render. Will override children if passed. */
-  content: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+  content?: string | object
   /** The children to render. Will be overridden by content prop if passed as well.  */
-  children: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
+  children?: React.ReactNode | object
```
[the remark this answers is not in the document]
guoda-puidokaite: This was mentioned as a future improvement. Added [here](https://github.com/cloudoperators/juno/issues/570).

guoda-puidokaite:
(**Improvement**)

For `AppShellProvider`, the `children` type is an `array` in `Storybook` but `React.ReactNode` in the code.

I think we could improve this for `AppShellProvider` (left old .JS code, right new .TSX code) :
Screenshot 2024-11-04 at 15 13 27

In other components, it's `any`, as Storybook doesn't have an appropriate type with a comment. 
Example of what we do in `Message`:

guoda-puidokaite (approved):
Small comment on types in Storybook, but idk if necessary to change.
Otherwise Thank You for the type changes! Looks great. 🚀💪🤪

packages/ui-components/src/components/AppShellProvider/AppShellProvider.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <StyleProvider theme={theme} stylesWrapper={shadowRoot ? "inline" : stylesWrapper}>
+        <PortalProvider>{children}</PortalProvider>
+      </StyleProvider>
+    </Wrapper>
+  )
+}
+
+export type AppShellStyleWrapper = "head" | "inline"
+
+interface WrapperProps {
```
barsukov, line 44: little comment from my side you can use it as absed props for AppShellProviderProps and extends it removing them from AppShellProviderProps

barsukov (commented):
I left one comment that kinda nice to fix before we merge.

packages/ui-components/src/components/AppShellProvider/AppShellProvider.component.tsx
gjaskiewicz: Done

guoda-puidokaite (approved):
LGTM too! 🚀 Thank You!

PR #548, opened 2024-10-23
chore(docs): adds a decision document for greenhouse/carbon

andypf:
This PR adds a decision document outlining the architecture for the next generation of the Greenhouse Dashboard. The document details key proposals for enhancing extensibility, scalability, and security while integrating extensions effectively.

Key Highlights

- Modular Design: Establishes a system for interchangeable modules and extensions.
- Routing and Loading: Describes isolated routing for extensions and a unified loading mechanism.
- Communication Protocols: Defines secure communication channels between the host and extensions.
- Security Measures: Outlines sandboxing, CORS, and authentication strategies.


docs/proposals/003_greenhouse.md
ArtieReus, line 1: Why is this file being removed instead of adapted? I thought this is the target of the proposal. Discuss, agree and change the proposals, not removing them without any agreement

ArtieReus (changes requested):

andypf:
Close this PR because of outdated content.

docs/proposals/003_greenhouse.md
edda: I agree with @ArtieReus. We should never delete files once they have been created. If they are superseded by another proposal then add this information to the beginning of the document but don't alter it in any other way.

PR #562, opened 2024-10-24
chore(ui): convert LoadingIndicator to Typescript

guoda-puidokaite:
# Summary

- Migrated the `LoadingIndicator` component to `TypeScript`, enhancing code quality, maintainability, and scalability.

# Changes Made

**TypeScript Migration**:

- Converted `LoadingIndicator` component files to `TypeScript`.

**Fixes/Improvements**:

- Allowed type `number` and `string` for `size`.
- Fixed edge cases for `size`, defaulting to size `96` when an _invalid_ number value (e.g. _empty_ string), `0` or a _negative_ value is passed.
- (Not related to ticket) Exported deprecated `StyleProvider` from ui-components/src `index` file.

**Testing**:

- Tested fixed edge cases.

**Cleanup:**

**Documentation:**

Screenshot 2024-10-25 at 09 14 09

**Release:**

# Related Issues

- https://github.com/cloudoperators/juno/issues/238

# Testing Instructions

1. Run `Storybook`
2. Run `LoadingIndicator` tests

# Checklist


packages/ui-components/src/components/LoadingIndicator/LoadingIndicator.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import React from "react"
+import Loading from "./loading-indicator.svg"
+
+export interface LoadingIndicatorProps {
+  /**
+   * The size of the LoadingIndicator in pixels.
+   */
+  size?: string
```
guoda-puidokaite, line 13: Few questions:
- Should we allow `size` as a `number` as well and account for that in the code? Is there a reason it's only a `string`?
- We expect the `size` to be passed as a `string`, without `px` included. What if the user does include `px` (or something else entirely) as part of the input? Is that fine? Or should we filter it out? Allowing only `number` type would really help here (if we indeed only want that)... but I understand it's not so easy to change it. Would just love to understand the constraints so we can cover edge cases here. 💪 Maybe I misunderstood something though. 

CC: @edda

edda: Those are good questions. I can't tell you why we made the choice of going with string here. I agree that going with number type instead would take care of a bunch of edge cases because people can only put numbers in, so there'll never be the edge case of people accidentally passing the unit together with the number.

The only thing I can think of is that the svg component requires the width/height to be a string but that can easily be solved by turning the passed number into a string. So I agree with you, let's make it number instead of string.

guoda-puidokaite: Perfect, Thank You so much. Noted it here: https://github.com/cloudoperators/juno/issues/234#issuecomment-2435175493 and I'll create a ticket for future improvements, as discussed. 🙏

barsukov (approved):
LGTM thank you for little fix in index

barsukov (changes requested):
```
ERROR in ./src/components/LoadingIndicator/LoadingIndicator.stories.tsx 5:0-46
Module not found: Error: Can't resolve './index.js' in '/Users/wowa/work/juno/packages/ui-components/src/components/LoadingIndicator'
```

guoda-puidokaite:
> ```
> ERROR in ./src/components/LoadingIndicator/LoadingIndicator.stories.tsx 5:0-46
> Module not found: Error: Can't resolve './index.js' in '/Users/wowa/work/juno/packages/ui-components/src/components/LoadingIndicator'
> ```

Ooops, somehow missed to commit that file... Pushed now.

barsukov (approved):

edda (approved):
Love the edge cases fixing 🥇 
I made two small changes to comments (took out CCloud reference, fixed color class name reference)

barsukov (approved):

PR #564, opened 2024-10-24
chore(carbon): refactor layout to match current dashboard

taymoor89:
# Summary

This PR refactors the side navigation and the general layout to match the current greenhouse dashboard.

# Changes Made

- Layout refactored
- Side navigation in new dashboard does not flicker
 
# Related Issues

- https://github.com/orgs/cloudoperators/projects/4/views/9?pane=issue&itemId=84503372&issue=cloudoperators%7Cjuno%7C552

# Screenshots (if applicable)

### Layout
  **Before:**
  Screenshot 2024-10-24 at 16 17 09

  **After:**
   Screenshot 2024-10-24 at 16 17 34

## Side Navigation
   **Greenhouse** greenhouse **Carbon** carbon

# Testing Instructions

1. `cd apps/carbon`
2. `npm i`
3. `npm run dev`

# Checklist


apps/carbon/index.html
```diff
@@ -1,13 +1,16 @@
...
+	<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+	<title>Carbon</title>
+</head>
+
+<body>
+	<div id="root"></div>
+	<script type="module" src="/src/index.tsx"></script>
+</body>
+
+</html>
```
andypf, line 16: missing new line. It looks like you’re using different formatting rules—an indent of 4 instead of 2.

apps/carbon/src/components/AppContent/AppContent.tsx
```diff
@@ -0,0 +1,33 @@
+/*
```
andypf, line 1: I’m not quite sure why you placed AppContent inside a folder also named AppContent? Are you expecting to add more AppContent-related components to this folder in the future?

andypf (commented):
Good job! 
just a few comments.

andypf:
It looks like this PR breaks the tests for Greenhouse. We should investigate and fix the issue.

apps/carbon/index.html
taymoor89: yes vscode formatting removed it and it did it only for html file. I'll check settings.

apps/carbon/src/components/AppContent/AppContent.tsx
taymoor89: I did it for the following reasons:
- **Consistency:** To ensure consistency with other components, which are placed inside their own directories.
- **Organization and Readability:** Based on my previous experience, organizing all major components and their subcomponents within directories keeps everything well-structured. In a TypeScript project, a component might need to expose some types, so having a types.ts file exported from index.ts is beneficial. In contrast, putting everything in one file makes the code less readable and harder to refactor.

apps/carbon/index.html
ArtieReus: > yes vscode formatting removed it and it did it only for html file. I'll check settings.

I've experienced the same issue. The HTML files are recognized as Nunjucks, which causes the last line to be removed. Changing the file format to HTML resolves it for me.

apps/carbon/src/components/AppContent/AppContent.tsx
ArtieReus: > I did it for the following reasons:
> 
> * **Consistency:** To ensure consistency with other components, which are placed inside their own directories.
> * **Organization and Readability:** Based on my previous experience, organizing all major components and their subcomponents within directories keeps everything well-structured. In a TypeScript project, a component might need to expose some types, so having a types.ts file exported from index.ts is beneficial. In contrast, putting everything in one file makes the code less readable and harder to refactor.

I like this pattern. We’ve previously used it to contextualize views with dependencies, but not for standalone components like AppContent or Providers. I'd be interested in applying it across to see how it turns out at the end.

edda: Agreed. I also like this pattern. Makes it cleaner.

apps/carbon/index.html
taymoor89: I discovered why new lines in HTML files are removed when formatted. Even though the `.prettierrc` file includes the rule `"endOfLine": "lf"` VS Code overrides this rule and formats HTML files using its default settings. If you’re facing this issue, here are a few possible solutions:

**- Apply Prettier Rules at the End:**
(I couldn’t get this to work)
**- Disable VS Code’s HTML Formatter in Workspace Settings:**
(I didn’t prefer this solution)
Screenshot 2024-10-25 at 15 00 27

**- Override the Default HTML Formatter with Prettier in Workspace Settings JSON:**
```"[html]": { "editor.defaultFormatter": "esbenp.prettier-vscode"},```
(This is the solution I chose)

**Review by andypf — DISMISSED** [review]
lgtm

File: apps/carbon/src/components/Extension/Extension.tsx [hunk 3]
```diff
@@ -0,0 +1,15 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from "react"
+import { ContentHeading } from "@cloudoperators/juno-ui-components"
+
+const Extension = ({ extension }: { extension: Extension }) => (
```
barsukov, line 9: Would it be deleted later right?

apps/carbon/src/index.scss
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
+// SPDX-License-Identifier: Apache-2.0
+
+body {
+	margin: 0;
```
barsukov, line 5: Styling issue?

barsukov (approved):
Few minor things you could fix in upcoming! good job 🥇

apps/carbon/src/components/Extension/Extension.tsx
taymoor89: no just the implementation will be different later.

apps/carbon/src/index.scss
taymoor89: good catch, fixed!

barsukov (approved):

TilmanHaupt (approved):

PR #540, opened 2024-10-21
chore(ui): bring the Modal back to fix the build of apps

barsukov:
# Summary

Very spontaneous fix and workaround for make still Modal and other PortalProvider depending components available until full of migration of typescript has not been done.

# Changes Made

- Modal and ModalFooter are back under deprecation folder
- Stories for deprecated modal deleted
- Change export of Modal to a deprecated one under index.js

# Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have made corresponding changes to the documentation (if applicable).
- [x] My changes generate no new warnings or errors.

PR #513, opened 2024-10-10
chore(ui): migrate PortalProvider to Typescript

gjaskiewicz:
# Summary

migrate PortalProvider to Typescript

# Changes Made

- migrate PortalProvider to Typescript
- temporarily use mock Message in tests

# Related Issues

- Issue 1: [link to issue](https://github.com/cloudoperators/juno/issues/245)

# Testing Instructions

1. `npm i`
2. `turbo test`

# Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have made corresponding changes to the documentation (if applicable).
- [ ] My changes generate no new warnings or errors.

PR #565, opened 2024-10-24
fix(ui): fixes z index stacking problems for PageHeader

edda:
# Summary

This PR fixes the z-index stacking problem introduced by the recent changes to `PortalProvider`. The main change introduced with the new `PortalProvider` was that every portaled component now renders in its own container with `z-index: 1` within the portal root element. The thought is to ensure a level playing field for all portaled components, especially when multiple portaled components are open at the same time to ensure that they stack properly. 

This has the side effect that components which aren't rendered in a portal will render above all portaled components if they have a z-index higher than 1. This happened with the `PageHeader` component. It had a `z-index` of 50 which means it was rendered above the `Panel`'s header, obscuring it.

# Changes Made

- Changed `PageHeader` `z-index` to 1 to ensure proper stacking with `Panel` and `Modal`
- Removed `z-index` from `Label`. It had a `z-index` of 10 which meant it would have been above `PageHeader` when scrolling down. We might have to revisit this again if it turns out that this breaks other edge cases with the `Label`
- Added `PortalProvider` decorator to the `Panel`, `PanelBody`, and `PanelFooter` stories to ensure the Portal renders in a portal root and gets the correct styling

# Related Issues

- https://github.com/cloudoperators/juno/pull/513
- https://github.com/cloudoperators/juno/pull/540

# Testing Instructions

1. cd apps/example
2. `npx turbo dev`
3. click on edit item to open panel
4. PanelHeader should be visible and not obscured by PageHeader

To test Panel stories check storybook

# Checklist


andypf (approved):
I clicked through the preview in the example app. The panel opens correctly over the header, and the modals are functioning as expected. Everything looks good from my perspective.

barsukov (approved):
I also checked through Preview all good 👍

andypf (approved):

Issue #570, opened 2024-10-28
[EPIC](ui): Inconsistencies and Improvement Collection


edda:
### Description

This epic serves as a central place to gather and document any inconsistencies, potential improvements, or refactoring ideas for the design system’s UI component library. The goal is to identify areas where the component library could be enhanced for better developer experience, API consistency, and overall usability. This epic will act as an initial collection hub, capturing insights and ideas as they arise.

### Objectives

**Identify Component Inconsistencies:**
Collect and document any inconsistencies found within the component APIs, naming conventions, or component behaviors to ensure a more cohesive library.

**Gather Refactoring Ideas:**
Record ideas for refactoring or restructuring components to improve the developer experience, maintainability, and performance of the design system.

### Collection

- [ ] Get an overview of all components' APIs and identify inconsistencies, e.g. different prop names for the same or similar attribute/behaviour, prop type inconsistencies across components

gjaskiewicz:
https://github.com/cloudoperators/juno/pull/589#discussion_r1836608217

PR #571, opened 2024-10-28
chore(ui): migrate Select component to TypeScript

barsukov:
# Summary

- Migrated the `Select`, `SelectOption`, `SelectDivider` components to `TypeScript`
- Deprecate components `Select`, `SelectOption`, 
- Utilised static typing.

# Changes Made

**TypeScript Migration**:

- Converted `Select`, `SelectOption`, `SelectDivider` component files to `TypeScript`.
- Introduced a deprecated `Select`, `SelectOption` component for `Javascript` files.
- Modify and simplify `SelectRow` component

**Refactoring**:

**Testing**:

- Removed `AppShellProvider` component from tests and stories, as not yet converted. But replace it with PortalProvider cause it should be a more accurate use case, since stories use PortalProvider instead.

**Release:**

- Generated a `changeset` file.

# Related Issues

- https://github.com/cloudoperators/juno/issues/249

# Testing Instructions

1. Run `Storybook`
2. Run `Select`, `SelectOption`, `SelectDivider` tests

# Checklist


guoda-puidokaite (commented):
Starting to review this component now. Here's an initial observation regarding imports. Not a blocking change at all, but would love to see the imports cleaned up and default directory exports utilised.

packages/ui-components/src/components/FormRow/FormRow.stories.js
```diff
@@ -5,7 +5,7 @@
 
 import React from "react"
 import PropTypes from "prop-types"
-import { FormRow } from "./index.js"
+import { FormRow } from "./FormRow.component"
```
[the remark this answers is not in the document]
barsukov: Not sure if I would like that sinse it is hard to keep track on... I do not know for me it is make sense to have more explicit import instead. First of all it would be much easy to Find and Replace staff

barsukov:
@guoda-puidokaite regarding export/import lets focus on migration make same way it iwas before then we would decide where we wanna go with it later with whoel project.

guoda-puidokaite:
Yah, not a blocking change for this PR as I mentioned. 🤪 But wanted to bring it up and see what you think. We can talk about it in the future. 🚀

packages/ui-components/src/components/FormRow/FormRow.stories.js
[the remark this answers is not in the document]
guoda-puidokaite: Well all imports are relative so (for me, at least) it makes sense to not have to read any path and think "Okay, both components/files are in the same directory" with `./`. 🙈 
What file it's in doesn't really matter for the import of every component. And if I was looking for the component, `index` is my single source of truth is for the path. 
Otherwise I don't see much use for the export in all of the `index` files. And as we see in this migration, the more specific we make the imports, the more component imports we need to change when only a file name changes e.g. `component.js` to `component.ts` or `Row.component` to `NewRow.component `etc.
But we can talk about it in the future, just wanted to bring up as I'm seeing it often during the Typescript migration.

packages/ui-components/src/components/Select/Select.component.tsx
```diff
@@ -0,0 +1,406 @@
...
+      | React.ReactElement<unknown, string | React.JSXElementConstructor<unknown>>
+      | Iterable<React.ReactNode>
+  ) => {
+    return !(typeof str === "string" && str.trim().length === 0)
+  }
+
+  const uniqueId = () => "juno-select-" + useId()
+
+  const theId = id || uniqueId()
+  const helptextId = "juno-select-helptext-" + useId()
```
guoda-puidokaite, line 186: We need two different IDs, right? Can't be reused? Resolve the comment if so, just wanna double check.

packages/ui-components/src/components/Select/Select.component.tsx
```diff
@@ -0,0 +1,406 @@
...
+    onChange && onChange(value || "")
+    onValueChange && onValueChange(value || "")
+  }
+
+  const portalContainerRef = usePortalRef()
+
+  // Headless-UI-Float Middleware
+  // In order to quickly debug middleware state, each parameter function can be passed the state to work with, e.g.
+  // autoPlacement((state) => console.log(state), ({crossAxis: true, [params…]}))
+  const middleware = [
```
guoda-puidokaite, line 245: Should we utilise `useMemo` to prevent recreating on each re-render?

packages/ui-components/src/components/Select/Select.component.tsx
```diff
@@ -0,0 +1,406 @@
...
+  truncateOptions = false,
+  valid = false,
+  value,
+  valueLabel,
+  variant = "default",
+  width = "full",
+  wrapperClassName = "",
+  ...props
+}: SelectProps) => {
+  const isNotEmptyString = (
```
guoda-puidokaite, line 172: Let's add a return type `boolean`. 

Imagine `!isNotEmptyString`. 🤪 But we don't have these cases anywhere, so naming is fine and makes the code more readable as is. 😅

packages/ui-components/src/components/SelectDivider/SelectDivider.component.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export interface SelectDividerProps {
+  /** add a custom className */
+  className?: string
+}
+
+/** A visual border/divider between SelectOptions or SelectGroups */
+export const SelectDivider = React.forwardRef<HTMLDivElement, SelectDividerProps>(
+  ({ className = "", ...props }, forwardedRef) => {
+    return <div className={`juno-select-divider ${dividerStyles} ${className}`} ref={forwardedRef} {...props}></div>
```
guoda-puidokaite, line 22: Could self close `/>`. Don't have to make this change, it's tiny. 🤪

packages/ui-components/src/components/SelectOption/SelectOption.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const selectContext = useContext(SelectContext)
+  const { truncateOptions: truncateOptions, addOptionValueAndLabel: addOptionValueAndLabel } = selectContext || {
+    truncateOptions: false,
+    addOptionValueAndLabel: () => {},
+  }
+
+  // send option metadata to the Select component via Context
+  useEffect(() => {
+    addOptionValueAndLabel(value, label, children)
+  }, [value, label, children])
```
guoda-puidokaite, line 78: Should `addOptionValueAndLabel` be a dependency here? Idk.

packages/ui-components/src/components/SelectOption/SelectOption.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+export const SelectOption = ({
+  children,
+  className = "",
+  disabled = false,
+  value = "",
+  label,
+  ...props
+}: SelectOptionProps) => {
+  const selectContext = useContext(SelectContext)
+  const { truncateOptions: truncateOptions, addOptionValueAndLabel: addOptionValueAndLabel } = selectContext || {
```
guoda-puidokaite, line 70: Redundant de-structuring names

```
const { truncateOptions, addOptionValueAndLabel } = selectContext || {
  truncateOptions: false,
  addOptionValueAndLabel: () => {},
};
```

packages/ui-components/src/components/SelectOption/SelectOption.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+          jn-min-h-[2.5rem]
+          ${optionStyles}
+          ${selected ? "juno-select-option-selected " + selectedOptionStyles : unselectedOptionStyles}
+          ${disabled ? "juno-select-option-disabled jn-opacity-50 jn-cursor-not-allowed" : ""}
+          ${truncateOptions ? "juno-select-option-truncate" : ""}
+          ${className}
+        `}
+          {...props}
+        >
+          {selected ? <Icon icon="check" size="18" className={`${selectedIconStyles}`} /> : ""}
```
guoda-puidokaite, line 95: No need for template literal, just `selectedIconStyles` directly.

packages/ui-components/src/components/SelectRow/SelectRow.stories.js
```diff
@@ -6,9 +6,8 @@
...
 import PropTypes from "prop-types"
 
-import { SelectRow } from "./index.js"
-import { SelectOption } from "../SelectOption/SelectOption.component"
-import { Default as DefaultSelectOptionStory } from "../SelectOption/SelectOption.stories"
+import SelectRow from "./SelectRow.component"
+import { SelectOption } from "../../deprecated_js/SelectOption/SelectOption.component"
 import { PortalProvider } from "../../deprecated_js/PortalProvider/PortalProvider.component"
 
 export default {
```
guoda-puidokaite, line 13: Do we've a future task to remove this component once users switch over? Let's discuss this offline, how you handle such deprecations in the team. 😊

guoda-puidokaite (changes requested):
Looks greattt! 🚀

Just had a few curious questions, code improvements and one type addition (most important requested change for ticket).

Thank You!

packages/ui-components/src/components/SelectDivider/SelectDivider.component.tsx
barsukov: ...

packages/ui-components/src/components/SelectRow/SelectRow.stories.js
barsukov: I mean yes for sure we have to remove them

barsukov:
I really think that some change requests are good improvements, but others I would not mix up with migration. Migration is migration, we need to move trhough as carefull and quick as possible. Then later on we could be more dedicated on refactoring topics, but it is nice that you point some staff out @guoda-puidokaite

guoda-puidokaite:
As mentioned, only the type addition `boolean` change is a requested change. Rest I'm asking questions about, maybe some topics are worth bringing to the improvements EPIC. And some not. 😉🚀

packages/ui-components/src/components/SelectOption/SelectOption.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+      render(
+        <Select>
+          <SelectOption value="Option 1" />
+        </Select>
+      )
+    )
+    expect(console.warn).toHaveBeenCalledTimes(1)
+    const toggle = screen.getByRole("button")
+    expect(toggle).toBeInTheDocument()
+    await waitFor(() => userEvent.click(toggle))
```
taymoor89, line 30: I understand that this is the migration so I'd not change it right now but something for the future that it is not recommended to use user interaction apis straight from  `userEvent` import we should rather create an instance `userEvent.setup()` and trigger events on that.
https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent

packages/ui-components/src/components/SelectOption/SelectOption.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+    expect(screen.getByRole("option")).toBeInTheDocument()
+    expect(screen.getByRole("option")).toHaveTextContent("Option 1")
+  })
+
+  test("renders a SelectOption with label as passed", async () => {
+    await waitFor(() =>
```
taymoor89, line 36: again no need to do anything right now but I'd say that is an unnecessary usage of `waitFor` in a lot of places which could easily increase the time it takes to run tests.

taymoor89 (dismissed):

guoda-puidokaite (dismissed):
Super niiice! One more down. Thank Youuu! 💪

packages/ui-components/src/components/SelectOption/SelectOption.test.tsx
barsukov: But anyway I guess you need an act wrapper, I think that was a leftovers but good point I would add them to our upcoming ui-componetns epic here https://github.com/cloudoperators/juno/issues/570

packages/ui-components/src/components/SelectOption/SelectOption.test.tsx
barsukov: Same, going to resolve them

packages/ui-components/src/components/Select/Select.component.tsx
edda: yes, this creates `id` attributes in the html markup. `id` attributes on html elements must be unique.

edda (changes requested):
👍 🚀 
Only one small change request.

edda (approved):

guoda-puidokaite (approved):
Thank You! 🚀💪✅

packages/ui-components/src/components/FormHint/FormHint.component.tsx
```diff
@@ -51,4 +51,6 @@ export interface FormHintProps {
   variant?: FormHintVariant
   /** Pass a custom className */
   className?: string
+  /** Pass a custom id */
+  id?: string
```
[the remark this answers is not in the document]
barsukov: I added it because it was used by other components but never defined in propTypes :)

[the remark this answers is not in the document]
barsukov: I will take care of this ticket https://github.com/cloudoperators/juno/issues/583

PR #575, opened 2024-10-30
chore(ui): migrate menu to typescript (#237) 

### LINKED ISSUE #237: [TASK](ui): Migrate Menu(All related) component to TypeScript
### Task Description

We need to convert Menu and all related components from @cloudoperators/juno-ui-components to TypeScript

**List of potential dependencies**

- Check files that used with Menu, MenuSection.

**Sub-tasks**

- [ ] change the name to .tsx of all related components
- [ ] Change the stories to .tsx and all related componetns
- [ ] fix all errors related to npm run typecheck
- [ ] Transfer all related tests to vitests from jest
- [ ] Check if npm run build is working and prove the working parts in storybook or example app
- [ ] Check that npm run lint is working correctly and fix all tslint errors

**Related Issues**
- #168 

**Additional Context**
Please check this online codemode editor to change the PropTypes to a proper Ts interfaces

https://github.com/mskelton/ratchet
Online editor: https://ratchet.mskelton.dev/

We need to be pretty concious about migration to ts cause it could create a potential breaking changes. Also if there will be too much changes at once maybe consider to split up the task again. Also some dependencies could be hidden and only discovered while executing the migration, could also lead to potential reconsidering the order of task execution.

franzheidl:
chore(ui): migrate menu to typescript (#237) 

Migrate `Menu`, `MenuItem`, and `MenuSection` components to Typescript.

Also some edge cases were identified related to the logic what element node type to render depending he props passed that we should resolve. Even thought the corresponding tests and stories with a `Menu` parent render as desired, when rendering standalone `MenuItem` stories we get differing results. Here's a danger that standalone `MenuItem` stories will disprove our claimed functionality, even though the functionality is correct when used IRL and the context intended.

`<MenuItem href="sap.com" />` should render an anchor `<a>` element.
`MenuItem onClick={someHandler} />` should render a button element.

`href` should take precedence over `onClick`:
`<MenuItem href="sap.com" onClick={someHandler} />` should render an anchor `<a>` element.

`<MenuItem label="some label" />` should render a plain div.

When `children` are passed, `label` should be ignored:
`<MenuItem label="some label"><button>Some Button</button></MenuItem>`should render `<div><button>Some Button</button></div>`

# Changes Made

`Menu`, `MenuItem`, and `MenuSection` components, stories, and tests migrated to TypeScript.

Tests and stories were updated and additional ones wwere added in order to double-check for rendering type edge cases.

In standalone `MenuItem` stories where element types were rendered incorrectly (they were always correctly rendered in the `Menu` stories), `onClick` was set to `null` explicitly in order to render the correct element in these stories, too.

# Related Issues

resolves #237


packages/ui-components/src/components/Menu/Menu.component.tsx
```diff
@@ -48,7 +50,7 @@ export const Menu = ({ children = null, variant = "normal", className = "", ...p
 					juno-menu 
 					juno-menu-${variant} 
 					${baseStyles} 
-					${variantStyles(variant)} 
+					${variant === "small" ? smallStyles : normalStyles}
```
barsukov, line 53: is my github looks crazy or iti s formating issue?

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
```diff
@@ -0,0 +1,111 @@
...
+  children = null,
+  className = "",
+  disabled = false,
+  href = "",
+  icon = null,
+  label = "",
+  onClick = undefined,
+  ...props
+}) => {
+  const menuContext = useContext(MenuContext) as MenuContextType | null
```
barsukov, line 77: WHy do we need a casting here not jus use useContext<MenuContextType | null> instead?

barsukov (changes requested):
Good good work, I just have few comments related to types

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
franzheidl: To be honest I may have copied this from somewhere, happy to change according to your suggestion.

packages/ui-components/src/components/Menu/Menu.component.tsx
franzheidl: Not exactly sure what you mean? I mainly tried to get rid of an extra function for this simple thing. In fact, we could get rid of using string literals here and use the tw-classes straightaway imho.

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
franzheidl: Hey Vova, I just checked, when doing as you suggested the `MenuItem` won't show the smaller styles according to the context (`small` and `normal` variants will look exactly the same)? 

I am not sure why this happens, I'm afraid this is beyond my current understanding of TypeScript 😬

barsukov: Ok let me know if you need any help on this one I will be back next week and we could have a look.

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const variant = menuContext?.variant ?? "normal"
+
+  // Determine which element to render:
+  const Element = href ? "a" : children ? "div" : onClick ? "button" : "div"
```
edda, line 81: Since `div` is the default if neither `href` nor `onClick` applies,  couldn't this be simplified to this?
```suggestion
  const Element = href ? "a" : onClick ? "button" : "div"
```

edda (changes requested):

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
guoda-puidokaite: That's what @franzheidl tried, but it was causing a bug so I fixed it by re-introducing https://github.com/cloudoperators/juno/pull/575/commits/48c526fbb65d2cd33e6bab0c3ca4a27b9cf1a6d8

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
barsukov: `
  const menuContext = useContext<MenuContextType | undefined>(MenuContext)`

franzheidl: Thank you Vova, your suggestion works fine! 🙏

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
franzheidl: This is the solution @guoda-puidokaite found to prevent storybook from wrapping their doing own onClick handler magic, resulting in the componentn always rendering a button. 

Since this is only an issue in the standalone `MenuItem` story (which will never occur IRL)  and we may want 
to get rid of these anyway, I'd suggest going with this solution for now.

barsukov:
Should we re-review again?

franzheidl:
@barsukov Hey, we have just found one more edge case we need to cover, will ping when ready. Already included your suggested change re using `MenuContext`, 🙏

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const handleClick = (_event: MouseEvent<HTMLButtonElement>) => {
+    onClick && onClick(_event)
+  }
+
+  return (
+    <HLMenu.Item
+      as={Element}
+      // conditionally pass a href attribute only if href was supplied to prevent errors from headless-ui internal checks:
+      {...(href && Element === "a" && !disabled && { href })}
```
guoda-puidokaite, line 91: Do we need something like this to ensure href is exclusively passed to anchor elements and prevent redundant prop assignments, causing conflict?

```suggestion
      {...(href && !disabled && Element === "a" ? { href } : {})}
```

packages/ui-components/src/components/MenuSection/MenuSection.component.tsx
```diff
@@ -20,7 +19,7 @@ const titleStyles = `
 	jn-p-2
 `
 /** Use MenuSection to structure and sub-divide MenuItems in a menu. All but the last MenuSection will render a visible divider at the bottom. Optionally, a MenuSection can have a title.*/
-export const MenuSection = ({ title = "", children = null, className = "", ...props }) => {
+export const MenuSection = ({ title = "", children = null, className = "", ...props }: MenuSectionProps) => {
```
guoda-puidokaite, line 22: Consistency
```suggestion
export const MenuSection: FC<MenuSectionProps> = ({ title = "", children = null, className = "", ...props }) => {
```

guoda-puidokaite (commented):
A few type improvements. Feel free to ignore any additional suggestions.

packages/ui-components/src/index.js
```diff
@@ -57,9 +57,9 @@ export { LoadingIndicator } from "./components/LoadingIndicator"
 export { MainContainer } from "./components/MainContainer/index.js"
 export { MainContainerInner } from "./components/MainContainerInner/index.js"
 export { MainTabs } from "./components/MainTabs/index.js"
-export { Menu } from "./components/Menu/index.js"
-export { MenuItem } from "./components/MenuItem/index.js"
-export { MenuSection } from "./components/MenuSection/index.js"
+export { Menu } from "./components/Menu/index.ts"
+export { MenuItem } from "./components/MenuItem/index.ts"
+export { MenuSection } from "./components/MenuSection/index.ts"
```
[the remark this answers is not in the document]
franzheidl: Happy to remove them.

guoda-puidokaite:
I dismissed @edda review, please let's double check her concerns were addressed and approve.

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
```diff
@@ -0,0 +1,111 @@
...
+  /** Whether the menu item is disabled */
+  disabled?: boolean
+  /** Pass the name of an icon the button should show. Can be any icon included with Juno. */
+  icon?: keyof typeof KnownIconsEnum
+  /** The label of the menu item. Will take precedence over children passed to the component. */
+  label?: string
+  /** Pass an href to the menu item. Will result in the menu item being rendered as an `<a>`. */
+  href?: string
+  /** Pass an onClick handler to the menu item. Will result in the menu item being rendered as a `<button>`. */
+  onClick?: (_event: MouseEvent<HTMLButtonElement>) => void
```
barsukov, line 56: I would suggest to put 
// eslint-disable-next-line no-unused-vars this instead of _event name variables with underscore

barsukov (changes requested):
One last little change from my side 👍

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
franzheidl: Done. I could remove the underscore, and now there is no unused-vars error anymore (there used to be one).

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
guoda-puidokaite: Regarding [this](https://github.com/cloudoperators/juno/pull/575#issuecomment-2457559924). Apologies, I just meant to request a type introduction, not logic change. Please test the changes and fight back if they break something. I'll put the logic back now, my mistake. 
I guess this shows though that the unit test introduced doesn't cover this case correctly too. It passed while issue persisted in last commit. 😅

guoda-puidokaite (approved):
Thanks, @franzheidl. Yes, I'd suggest to not change any of the logic. 
The Storybook issue was important to fix, since the suggested refactor introduced it. It didn't exist before. Therefore, I suggested to keep the logic original.
Regarding the issue re-introduced, see [comment](https://github.com/cloudoperators/juno/pull/575#discussion_r1830552148). I meant to only request a type introduction, not logic change. 🙈 My mistake. 
PR looks good from my side to go ahead. Thank You so much for the improvements! 🚀

franzheidl:
Thank YOU so much for your support, @guoda-puidokaite ! 🙏

The issue issue is almost driving me mad as I don't fully understand what happens there. But since the problem is in storybook, and the component behaves correctly under "real" circumstances, I think we better fix the problem where it is.

I would really like to move on with this, as I'll be re-thinking , re-structuring, and re-working the whole `Menu` cluster of components next (already on it). Migrating them to TS was/is just an intermediate step.

packages/ui-components/src/components/MenuItem/MenuItem.component.tsx
barsukov: but in a code it is stilll here did you push your changes?

franzheidl: Yes I did push, the underscore is removed in the code below – not sure what you are seeing?

Here's how the upstream code on this PR looks to me right now:

Screenshot 2024-11-06 at 10 44 36

franzheidl: Ah, sorry, yes, you are right, will remove in an instant (line 56).

franzheidl: Done. My apologies for the confusion.

guoda-puidokaite:
We should disable the `children` control (PR right), as was previously (live left) 🙈

Screenshot 2024-11-06 at 12 14 12

Screenshot 2024-11-06 at 12 17 30

franzheidl:
Well spotted @guoda-puidokaite! 

I have now updated the component so interactive styles (background change on hover, pointer cursor) are only applied when we render an interactive element as `MenuItem`, such as `<a>` and `<button>`, but not if we render a plain `<div>`.

Thank you for your attention to detail!

barsukov (approved):
LGTM

guoda-puidokaite (approved):
Thank Youuu!!! 💪

PR #577, opened 2024-10-30
feat(heureka): adds pagination for component instances on a service details panel

hodanoori:
# Summary
This PR adds pagination to the list of component instances for a selected service on the service details page.

# Changes Made

- Separated the GraphQL query for retrieving the list of component instances from the query for fetching service list data.
- Added a new ComponentInstanceList component to encapsulate the code for displaying component instances and handling its pagination.
- Included ComponentInstancesOfService in the global store slice’s view list to set initial query options, with a default `first: 8` parameter to display a maximum of 8 component instances per page, minimizing the need for scrolling within the panel.
- Renamed "name"to "ccrn" for  "Services," "Components," and "Support Group" entities in all graphql queries.

# Related Issues

- https://github.com/cloudoperators/heureka/issues/337

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


hodanoori:
- closes cloudoperators/heureka#337

apps/heureka/src/components/services/ComponentInstancesList.jsx
```diff
@@ -0,0 +1,123 @@
...
+      <Container py px={false}>
+        <Stack className="flex justify-end">
+          <Pagination
+            currentPage={currentPage}
+            isFirstPage={currentPage === 1}
+            isLastPage={currentPage === totalPages}
+            onPressNext={() => onPaginationChanged(currentPage + 1)}
+            onPressPrevious={() => onPaginationChanged(currentPage - 1)}
+            onKeyPress={(oKey) => {
+              if (oKey.code === "Enter") {
```
barsukov, line 109: Maybe extract to a naming function here what do you think?

apps/heureka/src/components/services/ComponentInstancesList.jsx
```diff
@@ -0,0 +1,123 @@
...
+      "ComponentInstancesOfService",
+      {
+        ...queryOptions,
+        filter: { serviceName: [serviceName] },
+      },
+    ],
+    enabled: !!queryClientFnReady && !!serviceName,
+  })
+
+  const items = useMemo(() => {
```
barsukov, line 39: I think you probably do not need use memo here do you?

apps/heureka/src/components/services/ComponentInstancesList.jsx
```diff
@@ -0,0 +1,123 @@
...
+    return data?.ComponentInstances?.edges || []
+  }, [data])
+
+  const pageInfo = useMemo(() => {
+    return data?.ComponentInstances?.pageInfo || {}
```
barsukov, line 44: why just do not use directly page info without useMemo?

barsukov (changes requested):
1) Maybe you do not need useMemo in places, but I am not sure please let me know what do you think 
2) Would be nice to write some tests if it necessary for the components that you introduce.

apps/heureka/src/components/services/ComponentInstancesList.jsx
andypf: I think the internal handling of useMemo consumes more performance resources than the benefits it provides in this case.

andypf (commented):
Great job, Hoda! Please be sure to address the comments.

apps/heureka/src/components/services/ComponentInstancesList.jsx
hodanoori: Do you mean to define different functions for each of above functions? I do not see a big advantage of that to be honest. So I leave it as it is, if you don't mind.

apps/heureka/src/components/services/ComponentInstancesList.jsx
hodanoori: I remove it, since I think the computation here is not complicate.

hodanoori:
> 1. Maybe you do not need useMemo in places, but I am not sure please let me know what do you think
> 2. Would be nice to write some tests if it necessary for the components that you introduce.

Thanks for your review. I removed useMemos and added a test for the newly added component. Please have a look here again.

apps/heureka/src/components/services/ComponentInstancesList.jsx
```diff
@@ -0,0 +1,114 @@
...
+      },
+    ],
+    enabled: !!queryClientFnReady && !!serviceCcrn,
+  })
+
+  const items = data?.ComponentInstances?.edges || []
+  const pageInfo = data?.ComponentInstances?.pageInfo || {}
+  const totalPages = pageInfo?.pages?.length || 0
+
+  const onPaginationChanged = (newPage) => {
```
ArtieReus, line 43: I would recommend to extract this logic to an own component including the Pagination component since you might probably use the pagination in each list and it would give you also the change to tested easier. What do you think?

apps/heureka/src/components/services/ComponentInstancesList.test.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+    const issueCountText = screen.queryByText("5")
+    expect(issueCountText).toBeTruthy() // Check if the issue count "5" is displayed
+
+    const severityText = screen.queryByText("High")
+    expect(severityText).toBeTruthy() // Check if the severity "High" is displayed
+  })
+
+  // Test: Simulates pagination interaction and verifies query updates
+  it("handles pagination changes", () => {
```
ArtieReus, line 105: With the component isolation you would just need to test the component and not each list.

ArtieReus (approved):
lgtm but please take my suggestions into account for the future.

apps/heureka/src/components/services/ComponentInstancesList.test.jsx
```diff
@@ -0,0 +1,134 @@
...
+  useGlobalsQueryClientFnReady: vi.fn(),
+  useGlobalsQueryOptions: vi.fn(),
+  useGlobalsActions: vi.fn(),
+}))
+
+vi.mock("../shared/LoadElement", () => ({
+  default: () => <div>Loading...</div>, // Mock loading element
+}))
+
+vi.mock("../shared/Helper", () => ({
```
barsukov, line 35: Why you also mock it this way?

barsukov (commented):
I have a few questions regarding mocking, especially if start mocking our own behaviour, it means that the component use this behaviour wrong, kinda a good sign to rethink the state lifecycle.

apps/heureka/src/components/services/ComponentInstancesList.test.jsx
hodanoori: Cause I did not want to test the severityString, highestSeverity functions from helper inside this component. There is another test for helper.

apps/heureka/src/components/services/ComponentInstancesList.test.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+vi.mock("../StoreProvider", () => ({
+  // Mock StoreProvider to just render children
+  StoreProvider: ({ children }) => <div>{children}</div>,
+  useGlobalsQueryClientFnReady: vi.fn(),
+  useGlobalsQueryOptions: vi.fn(),
+  useGlobalsActions: vi.fn(),
+}))
+
+vi.mock("../shared/LoadElement", () => ({
```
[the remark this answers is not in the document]
hodanoori: I use this in my component to show loading busy when data isLoading but I did not want to test it here. That's why.

hodanoori:
> I have a few questions regarding mocking, especially if start mocking our own behaviour, it means that the component use this behaviour wrong, kinda a good sign to rethink the state lifecycle.

I do not get your point, Those mocks are belonging to other components which has their own tests.

hodanoori:
> I have a few questions regarding mocking, especially if start mocking our own behaviour, it means that the component use this behaviour wrong, kinda a good sign to rethink the state lifecycle.

I separated the pagination component and added a test for that component. Please have a look there again.

barsukov (approved):
LGTM

andypf (approved):
LGTM

PR #578, opened 2024-10-31
refactor(url-state-provider): moved to typescript

### LINKED ISSUE #163: [TASK]: Convert @cloudoperators/juno-url-state-provider to TypeScript
### Task Description

We need to convert  @cloudoperators/juno-state-provider to TypeScript

**Sub-tasks**

- [ ] Integrate Lint with typescript config(look to juno-communicator package as example)
- [ ] Integrate tsconfig.json from config package
- [ ] Integrate vitest for testing
- [ ] Integrate tsup for bundling things up
- [ ] Delete old rollup/esbuild/babel
- [ ] Makes tests green integrate in example app.
- [ ] Update readme

**Related Issues**
- #119 

**Additional Context**
Look to already existing configuration in juno-communicator

TilmanHaupt:
# Summary

This pull request refactors the app url-state-provider and all associated tests from JavaScript to TypeScript. The transition aims to improve type safety, enhance code maintainability, and leverage TypeScript's features for better development experience.

# Related Issues

closes #163 

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


packages/url-state-provider/src/index.ts
```diff
@@ -6,23 +6,36 @@
...
+const onGlobalChangeListeners: any[] = []
 
 /**
  * Encode json data using json-url or lz-string. It automatically detects the best encoding.
  * @param {JSON} json data to be encoded
  * @param {Object} options options for the encoding. Possible values: mode: "auto" or "humanize"
  * @returns encoded string
  */
-function encode(json, options = {}) {
+function encode(json: object, options: EncodeOptions = {}) {
```
andypf, line 38: Wouldn’t it be better to make the options parameter optional and define a return value for encode?

```typescript
function encode(json: object, options?: EncodeOptions): string {
  // Set options to an empty object if it's undefined
  options = options || {};
  // ...
```

This way, you can call encode without passing options, allowing the tests to remain unchanged.

packages/url-state-provider/src/index.ts
```diff
@@ -58,19 +71,19 @@ function decode(string) {
...
 }
 
 /**
- * find search param by key and convert it to json
- * @param {string} searchKey
+ * find search param by key (regex) and convert it to json
  * @returns json
  */
-function URLToState() {
+function URLToState(): { [key: string]: any } {
```
andypf, line 86: I’m wondering if we could replace the any type with a more specific one, like this:

```typescript
type DecodedState = { [key: string]: string | number | boolean | any[] | null };

function URLToState(): DecodedState {
  // ...
```

This change would enhance type safety and provide better clarity regarding the structure of the returned object.
What do you think?

packages/url-state-provider/src/index.ts
```diff
@@ -122,17 +135,17 @@ function currentState(stateID) {
  * @param {JSON} state
  * @returns new url string
  */
-function updateState(stateID, state, options = {}) {
+function updateState(stateID: string, state: object, options: UpdateStateOptions = {}) {
   // get current state from URL
-  const newState = URLToState()
+  const newState: { [key: string]: any } = URLToState()
```
andypf, line 140: see my last comment

packages/url-state-provider/src/index.ts
```diff
@@ -152,16 +165,15 @@ function updateState(stateID, state, options = {}) {
  * changes to the method.
  * @param {boolean} historyOptions.replace - If true it replaces the last state in the window history (default false).
  */
-function updateStateAndHistory(stateID, state, merge, historyOptions) {
-  historyOptions = historyOptions || {}
+function updateStateAndHistory(stateID: string, state: object, merge: boolean, historyOptions: HistoryOptions = {}) {
```
andypf, line 168: Instead of casting you could define the type of newUrl, or what do you think?

```typescript
function updateStateAndHistory(stateID: string, state: object, merge: boolean, historyOptions: HistoryOptions = {}) {
  const newUrl: string = updateState(stateID, state, { merge }); // Ensure newUrl is typed as a string
  const historyState = historyOptions.state || "";
  const historyTitle = historyOptions.title || "";

  if (historyOptions?.replace) {
    window.history.replaceState(historyState, historyTitle, newUrl);
  } else {
    window.history.pushState(historyState, historyTitle, newUrl);
  }
}
```

packages/url-state-provider/src/index.ts
```diff
@@ -255,7 +267,7 @@ function informListener(stateID, newState, oldState) {
   listener(newState)
 }
 
-function onGlobalChange(callback) {
+function onGlobalChange(callback: any): any {
```
andypf, line 270: Same here, may be there is a way to avoid `any`?

andypf (commented):
Great job, Tilman! 👍  
I have a few comments; I think we can further improve the code by using more specific types, for example.

packages/url-state-provider/src/index.ts
TilmanHaupt: This parts are more your code so i was a bit less specific, because I understand less. So I guess you are right - in the end i can take the same type like in superstate.

packages/url-state-provider/src/index.ts
TilmanHaupt: Screenshot 2024-11-05 at 11 36 24

TilmanHaupt: Do you know - why

PR #579, opened 2024-11-01
fix(ui): adds high z-index to portal root

edda:
# Summary

This PR tries to improve upon the previous fix for the z-index stacking problem introduced in the recent refactoring of `PortalProvider`. While the previous fix in https://github.com/cloudoperators/juno/pull/565 solved the issue for Juno apps we still had potential z-index stacking issues if a Juno app was embedded into a host app with its own z-index stack. This is because the solution from the previous PR aimed to standardize all z-index on the root level to `1` and then added stacks above this in the portaled components as necessary. 

In the cases where a Juno app is embedded in a host app with z-indexes higher than `1` the whole thing would break.

This PR aims to fix this by setting a very high z-index (`9999`) for the portal root div which acts as the wrapper for all portals. So as long as the host app's z-index is at max `9999` the z-stacking will be correct.

# Changes Made

- Added `z-index: 9999` to portal root div in `PortalProvider`
- Undid `z-index` change to `PageHeader` and `Label` from PR#565 since the new portal root `z-index` ensure proper stacking above these components for portaled components

# Related Issues

- https://github.com/cloudoperators/juno/pull/565

# Testing Instructions

1. start exampleapp or supernova in this branch
2. open a Panel --> it should be stacked above the PageHeader
3. in Supernova scroll down so that the filter toolbar gets scrolled under PageHeader --> the Select's Label should not be visible

# Checklist

### DIFF
--- a/.changeset/tiny-cooks-try.md
+++ b/.changeset/tiny-cooks-try.md
@@ -1,0 +1,5 @@
+---
+"@cloudoperators/juno-ui-components": patch
+---
+
+adds very high z-index (9999) to portal root container
--- a/packages/ui-components/src/components/Label/Label.component.tsx
+++ b/packages/ui-components/src/components/Label/Label.component.tsx
@@ -13,6 +13,7 @@ const labelstyles = `
   jn-transition-all 
   jn-duration-100 
   jn-ease-in-out
+  jn-z-10
 `
 
 const floatingStyles = `
--- a/packages/ui-components/src/components/PageHeader/PageHeader.component.tsx
+++ b/packages/ui-components/src/components/PageHeader/PageHeader.component.tsx
@@ -13,7 +13,7 @@ const pageHeaderStyles = `
   jn-top-0
   jn-px-6
   jn-py-3
-  jn-z-[1]
+  jn-z-50
 `
 
 const pageHeaderInnerStyles = `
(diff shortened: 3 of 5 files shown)


hodanoori (approved):
Works as described in both Exampleapp and Supernova, so I approve it 👍

ArtieReus:
- Tested changes in elektra:
Screenshot 2024-11-01 at 16 45 46
- Tested changes in exampleapp
Screenshot 2024-11-01 at 16 53 38
- supernova
Screenshot 2024-11-01 at 17 00 07
all look good

ArtieReus (approved):
as I said in the comment. LGTM

andypf:
We could also define a property in the AppShellProvider and other components like AppShell to set the starting z-index. For example, we could specify startZIndex={60}. This would allow us to handle the z-index as usual (e.g., starting from 1 as in your last fix) while covering the Elektra case by setting the starting z-index higher than 50, such as 60.

edda:
> We could also define a property in the AppShellProvider and other components like AppShell to set the starting z-index. For example, we could specify startZIndex={60}. This would allow us to handle the z-index as usual (e.g., starting from 1 as in your last fix) while covering the Elektra case by setting the starting z-index higher than 50, such as 60.

I like this idea. But it needs some more thought how exactly we want to do it. I'll add it to our collection epic for future improvements 🙏

PR #599, opened 2024-11-09
chore(ui): migrate NativeSelect, NativeSelectOption and NativeSelectOptionGroup components to TypeScript

guoda-puidokaite:
# Summary

- Migrated `NativeSelect`, `NativeSelectOption` and `NativeSelectOptionGroup` components to `TypeScript`, enhancing code quality, maintainability, and scalability.
- Refactored the components for cleaner, more maintainable code.
- Kept `NativeSelect` and `NativeSelectOption` components for use in `Javascript` files.
- **Note** bugs and improvements found: https://github.com/cloudoperators/juno/issues/240#issuecomment-2468482674 for [improvement epic](https://github.com/cloudoperators/juno/issues/570).

# Changes Made

**TypeScript Migration**:

- Converted `NativeSelect`, `NativeSelectOption` and `NativeSelectOptionGroup` components files to `TypeScript`.
- Moved deprecated `NativeSelect` and `NativeSelectOption` components to `/deprecated_js` for `Javascript` files.
- Exported migrated components instead of deprecated.
- Ensured additional props are valid intrinsic props.

**Refactoring**:

**Testing**:

- Added edge cases:
  - Native Select
    - Without `children`
    - Multiple `SelectOptionGroup` components
    - Without `value` or `label`
  - Native Select Option
    - No `value` or `label` provided
    - `Number` type for `value` and `label`
  - Native Select Option Group
    - `Empty` label prop
    - undefined` children
    - Multiple `children` of different `types`
    - Empty component
    - Invalid props
- Organised tests by what they check for better readability and maintenance.
- Verified components behaviour using `Storybook`.

**Documentation:**

**Release:**

- Generated a `changeset` file.

# Related Issues

- Closes https://github.com/cloudoperators/juno/issues/240

# Testing Instructions

1. Check `Storybook`

# Checklist


packages/ui-components/src/components/NativeSelect/NativeSelect.component.tsx
```diff
@@ -0,0 +1,254 @@
...
+  )
+}
+
+/**
+ * A basic, uncontrolled native HTML select component.
+ * Takes NativeSelectOption and NativeSelectOptionGroup children as options.
+ * Handles states such as loading, valid, invalid, and error.
+ */
+export const NativeSelect: FC<NativeSelectProps> = ({
+  name = "Unnamed Select",
```
andypf, line 195: In the original version, the value was set to null. Is there a specific reason for changing it to a default String value?

andypf (commented):
Overall, this looks great to me! I’ve added just a few comments for consideration.

packages/ui-components/src/components/NativeSelect/NativeSelect.component.tsx
guoda-puidokaite: We didn't assign it as a default value, [but in our code](https://github.com/cloudoperators/juno/blob/eea5de67a22b0429506040285cde1d1e2d35739d/packages/ui-components/src/components/NativeSelect/NativeSelect.component.js#L194), it's technically the default value. Seemed correct to let the user know explicitly. What do you think?

packages/ui-components/src/components/NativeSelectOption/NativeSelectOption.component.tsx
```diff
@@ -0,0 +1,49 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from "react"
+
+export interface NativeSelectOptionProps extends Omit<React.OptionHTMLAttributes<HTMLOptionElement>, "label"> {
```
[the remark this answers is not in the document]
guoda-puidokaite: Ah, good point. `label` was the only one causing a conflict in the code base, which is why I omitted. Omitting `value` isn't strictly necessary, but definitely creates stronger type safety. Let's do it. Thank You! 🚀

packages/ui-components/src/components/NativeSelect/NativeSelect.component.tsx
andypf: The name is sent as a property in forms. While it’s unlikely that we’ll use forms in the React context, if we do, the name will be set, which could potentially confuse developers. If this value is null, the selected option will not be included when the form is submitted. This is the expected behavior, and I wouldn’t recommend changing it.

image

guoda-puidokaite: Will definitely change it. 🙈 Just really wanna understand what I'm missing first.
In old code, we initially set `name` as `null` by default. [But then in our code](https://github.com/cloudoperators/juno/blob/eea5de67a22b0429506040285cde1d1e2d35739d/packages/ui-components/src/components/NativeSelect/NativeSelect.component.js#L194), if `name` isn't set, we always set it to  `Unnamed Select` (as a **hidden default**).

I just wanted to make the default explicit so it indeed doesn't confuse the user.
But maybe I'm missing something in between, how setting the default initially changes the overall behaviour? 🙈

**Side note:**
Now that you bring it up though... Maybe we should remove the default `name` value from code all together? To make sure the option is only included in form submissions if explicitly set? Could add to future improvements.

andypf: Let's do improvements after migration.

andypf (approved):
Great Job!

packages/ui-components/src/components/NativeSelect/NativeSelect.component.tsx
guoda-puidokaite: Added https://github.com/cloudoperators/juno/issues/240#issuecomment-2468482674

packages/ui-components/src/components/NativeSelect/NativeSelect.component.tsx
```diff
@@ -0,0 +1,254 @@
...
+  onClick,
+  wrapperClassName = "",
+  ...props
+}) => {
+  const [isLoading, setIsLoading] = useState(loading)
+  const [isValid, setIsValid] = useState(valid)
+  const [isInvalid, setIsInvalid] = useState(invalid)
+  const [hasError, setHasError] = useState(error)
+
+  useEffect(() => setIsLoading(loading), [loading])
```
barsukov, line 214: All of this useEffects are so unnecessary I would really put it into consideration of future improvements.

packages/ui-components/src/components/NativeSelect/NativeSelect.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+  ],
+}
+
+export const LoadingNativeSelect = Template.bind({})
+LoadingNativeSelect.args = {
+  loading: true,
+  children: [],
+}
+
+export const NativeSelectWithError = Template.bind({})
```
barsukov, line 90: Do we need to export all of them? I think to reuse them in other components right? also nice point of improvements like Epic for grouping them all together in one...

packages/ui-components/src/components/NativeSelect/NativeSelect.component.tsx
guoda-puidokaite: Done: https://github.com/cloudoperators/juno/issues/570 (by you). But I added this one as an example.

packages/ui-components/src/components/NativeSelectOption/NativeSelectOption.component.tsx
```diff
@@ -0,0 +1,50 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from "react"
+
+export interface NativeSelectOptionProps
+  extends Omit<React.OptionHTMLAttributes<HTMLOptionElement>, "label" | "value"> {
```
barsukov, line 9: Super!

packages/ui-components/src/components/NativeSelectOptionGroup/NativeSelectOptionGroup.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+   * Additional CSS classes to apply to the option group for custom styling.
+   */
+  className?: string
+
+  /**
+   * Elements to be rendered within the option group.
+   * This can be any React node or a collection of React nodes.
+   * Typically, these should be NativeSelectOption components.
+   */
+  children?: React.ReactNode
```
barsukov, line 31: Should you do here maybe same as for combobox PR like restrict element with a generic props??

barsukov (commented):
I leave it as a comment, all good few little comments left do not wanna block 🥇

packages/ui-components/src/components/NativeSelectOptionGroup/NativeSelectOptionGroup.component.tsx
guoda-puidokaite: Didn't get around to it, thanks for catching that. I really like the approach if we expect specific children. 💪

andypf (dismissed):
👍

ArtieReus:
I hope this hasn’t been mentioned already, but I just noticed that the default value for `NativeSelect` is set to `Unnamed Select` instead of an empty string. Is this intentional?
Screenshot 2024-11-13 at 14 26 01

ArtieReus:
> I hope this hasn’t been mentioned already, but I just noticed that the default value for `NativeSelect` is set to `Unnamed Select` instead of an empty string. Is this intentional? Screenshot 2024-11-13 at 14 26 01

sorry... i just have been pointed that it is already clarified.
https://github.com/cloudoperators/juno/pull/599#discussion_r1836462784

ArtieReus:
Should the label not being defaulted to empty string?
Screenshot 2024-11-13 at 14 41 02

packages/ui-components/src/components/NativeSelectOptionGroup/NativeSelectOptionGroup.component.tsx
guoda-puidokaite: Reverted this change in case it breaks things for how apps are using children now. 
Improvement added to [epic](https://github.com/cloudoperators/juno/issues/570) under "For some elements, we could make sure children are only elements (and props) we expect, if parent only works with those children".

guoda-puidokaite:
> Should the label not being defaulted to empty string? Screenshot 2024-11-13 at 14 41 02

Thanks for the review! [We're not currently doing this](https://github.com/cloudoperators/juno/blob/main/packages/ui-components/src/components/NativeSelectOptionGroup/NativeSelectOptionGroup.component.js). But I definitely think it's a good idea to introduce it, making sure it's always safely handled as a `string`. Good catch.

packages/ui-components/src/components/NativeSelect/NativeSelect.component.tsx
```diff
@@ -0,0 +1,254 @@
...
+    jn-select-none
+    jn-cursor-not-allowed
+  `,
+  loadingSpinner: `jn-ml-auto jn-mr-auto`,
+  errorIcon: `jn-ml-auto jn-mr-auto`,
+  paddingIconRight: `jn-pr-[3.75rem]`,
+  paddingDefaultRight: `jn-pr-9`,
+}
+
+export interface NativeSelectProps {
```
ArtieReus: should not be extended here from something like `React.HTMLAttributes<HTMLDivElement>`?

guoda-puidokaite: Already committed, thanks for the catch. 🚀

packages/ui-components/src/components/NativeSelectOptionGroup/NativeSelectOptionGroup.component.tsx
andypf: ok

PR #603, opened 2024-11-13
fix(heureka): corrects URL state handling for filter property

hodanoori:
# Summary

Fixed an issue where activeFilters were not immediately applied after URL updates, requiring additional actions to trigger synchronization. Refactored useUrlState by updating useEffect dependencies to ensure real-time syncing of URL state and filters when changes occur.
Additionally, restored the missing activeFilters prop in FilterSelect within the Filters control.

Furthermore, the detailsFor prop is now handled correctly, ensuring that when a detail panel is closed, the corresponding detailsFor information is removed from the URL, maintaining a clean and accurate URL state.

# Changes Made

- Updated useEffect dependencies in useUrlState for immediate filter synchronization.
- Restored the activeFilters prop in FilterSelect within the Filters control.
- Added logic to remove detailsFor and associated details from the URL when the panel is closed.

# Related Issues

- https://github.com/cloudoperators/heureka/issues/375

# Testing Instructions

- Apply filters on different views in the app.
- Verify that the selected filters for each view are reflected in the URL.

For example, the following part of URL should display the `Support group ccrn = compute-storage-api` filter on the `Services` view:
f:(Components:(:),IssueMatches:(:),Services:(supportGroupCcrn:(compute~Fstorage~Fapi)))

1. `npm i`
2. `npm run TASK`

# Checklist


andypf (approved):
LGTM 👍

apps/heureka/src/hooks/useUrlState.js
```diff
@@ -57,7 +63,7 @@ const useUrlState = () => {
     }
 
     setIsURLRead(true)
-  }, [isURLRead])
+  }, [isURLRead, setActiveView, setFiltersFromURL, setShowPanel, setServiceDetail, setIssueDetail])
```
ArtieReus, line 66: I was a bit confused here because I have understood that if the setter functions are stable and don’t change between renders, you can safely omit them from the dependencies BUT it doesn´t hurt.

apps/heureka/src/hooks/useUrlState.js
```diff
@@ -74,7 +80,17 @@ const useUrlState = () => {
     if (JSON.stringify(updatedState) !== JSON.stringify(currentState)) {
       urlStateManager.push(updatedState)
     }
-  }, [loggedIn, detailsFor, activeView])
+  }, [
+    loggedIn,
```
ArtieReus, line 84: Are you still using loggedIn? I thought all authentication-related code was removed.

apps/heureka/src/hooks/useUrlState.js
```diff
@@ -74,7 +80,17 @@ const useUrlState = () => {
...
+    detailsFor,
+    activeView,
+    serviceActiveFilters,
```
ArtieReus, line 87: This useEffect looks confusing to me since there is no relation from the dependency array to the code being used inside in the hook. This could be improved in a new PR. Would you create a task for it?

apps/heureka/src/hooks/useUrlState.js
hodanoori: No, actually they should be listed here, as otherwise the useEffect does not affect on applying filter changes right after selecting them into URL.

apps/heureka/src/hooks/useUrlState.js
hodanoori: No, loggedIn should be deleted. Thanks for the reminder.

apps/heureka/src/hooks/useUrlState.js
hodanoori: Let's discuss this, since as I said these dependencies are taking care of updating the filter props in the URL. The task that I mentioned in this PR, is in fact for solving this problem, so we do not need a new task, I suppose.

andypf (approved):
Now I see that the filters are correctly applied to the URL. Looks great to me—good job!

ArtieReus (commented):
Hi! I started checking the preview, and here’s what I noticed so far: the current page in pagination isn't being saved in the URL. For example, if I'm on page 3/3 in the Services tab and click on a service, the details panel opens correctly. However, when I reload the page, while the panel is correctly populated, the pagination resets to page 1/3, breaking the connection to the details shown in the panel. Could this behavior be related to the recent improvements?

ArtieReus (commented):
I'm experiencing the same issue with the search term. It seems that both pagination and the search term are not being reflected in the URL, correct?

ArtieReus (commented):
By clicking I have noticed that the description of an IssueMatch in the details panel stays in Loading state for ever. Is this an Issue and do you want to create a ticket for it?
Screenshot 2024-11-15 at 09 16 05

apps/heureka/src/hooks/useUrlState.js
```diff
@@ -57,24 +65,35 @@ const useUrlState = () => {
...
 
   // Sync URL with the desired states
   useEffect(() => {
     if (!isURLRead) return
 
     const updatedState = {
       [constants.ACTIVE_VIEW]: activeView, // Include active view
-      ...syncFiltersWithURL(),
-      ...syncDetailsWithURL(),
+      ...syncFiltersWithURL(issueMatchesActiveFilters, serviceActiveFilters, componentActiveFilters),
```
ArtieReus, line 76: Much better, I like it :clap:

apps/heureka/src/lib/slices/createGlobalsSlice.js
```diff
@@ -124,13 +124,19 @@ const createGlobalsSlice = (set, get, options) => ({
         ),
 
       // Sync details state with the URL
-      syncDetailsWithURL: () => {
-        const state = get().globals
+      syncDetailsWithURL: (serviceDetail, issueDetail, panelType) => {
```
ArtieReus, line 127: Here’s a proposal for the future: since this function uses the globals internally without modifying anything in the store, it might be better to refactor it into a helper function where globals are passed as a parameter. What do you think?

hodanoori:
> Hi! I started checking the preview, and here’s what I noticed so far: the current page in pagination isn't being saved in the URL. For example, if I'm on page 3/3 in the Services tab and click on a service, the details panel opens correctly. However, when I reload the page, while the panel is correctly populated, the pagination resets to page 1/3, breaking the connection to the details shown in the panel. Could this behavior be related to the recent improvements?

Hi Arturo, nice catch, yes, the pagination is not correctly kept in the URL. But I will add the check for pagination and search in url state in a separate PR, if it's fine for you.

hodanoori:
> I'm experiencing the same issue with the search term. It seems that both pagination and the search term are not being reflected in the URL, correct?

Correct, I fix the search along with pagination in next PR.

hodanoori:
> By clicking I have noticed that the description of an IssueMatch in the details panel stays in Loading state for ever. Is this an Issue and do you want to create a ticket for it? Screenshot 2024-11-15 at 09 16 05

This is known to me, I fix this issue also in a separate PR.

PR #601, opened 2024-11-12
chore(ui): migrate Form, FormSection and FormRow components to Typescript and update FormHint

guoda-puidokaite:
# Summary

- Migrated `Form`, `FormSection` and `FormRow` components to `TypeScript`, enhancing code quality, maintainability, and scalability.
- Utilised static typing to minimize runtime errors and improve the development experience.
- Refactored the components for cleaner, more maintainable code. Including `FormHint` (which was already migrated).
- Kept `FormRow` component for use in `Javascript` files and deleted others.
- **Note** improvements found: https://github.com/cloudoperators/juno/issues/228#issuecomment-2470461643 for [improvement epic](https://github.com/cloudoperators/juno/issues/570).
- **Another Note**: Added back removed storybook components as in [this comment](https://github.com/cloudoperators/juno/issues/228#issuecomment-2470485849)

# Changes Made

**TypeScript Migration**:

- Converted `Form`, `FormSection` and `FormRow` components files to `TypeScript`.
- Added comprehensive type annotations.
- Moved deprecated `FormRow` component to `/deprecated_js` for `Javascript` files and deleted the others.
- Exported migrated components instead of deprecated.
- Ensured additional props are valid intrinsic props.

**Refactoring**:

- Made necessary adjustments to address TypeScript-detected issues.
- Further refactored for cleaner and more understandable code (including `FormHint`).

**Testing**:

- Added edge cases:
  - Form
    - Render without any props
    - Render with 0 as title
    - Render with falsy values as title
    - Applies default styles
    - Render multiple children
    - Render with no children
  - FormSection
    - Render without any props
    - Render with null children
    - Render with undefined children
    - Render with mixed children types
    - Render with an empty string as title
    - Render with a long string as title
    - Render with special characters in title
    - Render h1 element only if title is provided
    - Render section HTML element with classnames
  - FormRow
    - Render without any props
    - Render with null children
    - Render with undefined children
    - Adds and handles additional props correctly
    - Applies default styles
    - Render multiple children
    - Render with no children
  - FormHint
    - Handles an empty string as text
    - Handles null text
    - Handles undefined text
    - Renders very long text
- Organised tests by what they check for better readability and maintenance.
- Verified components behaviour using `Storybook`.
- Updated tests to use `vitest`.

**Documentation:**

- Updated `Storybook` documentation.

**Release:**

- Generated a `changeset` file.

# Related Issues

- Closes https://github.com/cloudoperators/juno/issues/228

# Testing Instructions

[...]

PR #607, opened 2024-11-14
feat(ui): adds HeaderContainer component

MartinS-git:
# Summary

To enable <TopNavigation> to behave like <PageHeader>, a container component was created: <HeaderContainer>, which is placed in the <AppShell>. The result is that <TopNavigation> is now also sticky and always remains at the top when scrolling. In addition, the width of both components behaves in the same way as the content. This means that they are now also bordered and centered in width.

# Changes Made

- Created new component <HeaderContainer>
- Changes on <AppShell>: <HeaderContainer> added. In addition, there was a command that gave the MainContainer a margin to the left. As this is no longer required, it has been deleted
- Example App: TopNavigation added 

# Related Issues

There is no ticket

# Screenshots (if applicable)

image

# Testing Instructions

The behavior can only be viewed in the example app
1. `npm i`
2. `npm run clean:cache `
3. `npm run pkg @cloudoperators/juno-app-example`

Component in Storybook
1. `npm i`
2. `npm run storybook -w @cloudoperators/juno-ui-components `

# Checklist


packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -41,7 +42,7 @@ export const AppShell = ({
       {contentHeading || ""}
       {embedded ? (
         <>
-          {topNavigation && topNavigation}
+          <HeaderContainer>{topNavigation && topNavigation}</HeaderContainer>
```
andypf, line 45: Please change `{topNavigation && topNavigation}` to just `{topNavigation}` for simplicity

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -55,22 +56,23 @@ export const AppShell = ({
...
       ) : (
         <>
-          {pageHeader && (typeof pageHeader === "string" || pageHeader instanceof String) ? (
-            <PageHeader heading={pageHeader} />
-          ) : (
-            pageHeader
-          )}
-          {topNavigation && topNavigation}
-          {/* Wrap everything except page header and footer and navigations in a main container. Add top margin to MainContainerInner as we are not in embedded mode here. */}
+          <HeaderContainer fullWidth={fullWidthContent === true ? true : false} className="">
```
andypf, line 59: Please change `{fullWidthContent === true ? true : false}` to just `{fullWidthContent === true}` for simplicity

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -55,22 +56,23 @@ export const AppShell = ({
...
+            {pageHeader && (typeof pageHeader === "string" || pageHeader instanceof String) ? (
+              <PageHeader heading={pageHeader} />
+            ) : (
+              pageHeader
+            )}
+            {topNavigation && topNavigation}
```
andypf, line 65: ` {topNavigation && topNavigation}` -> ` {topNavigation}`

packages/ui-components/src/components/HeaderContainer/HeaderContainer.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+  fullWidth = false,
+  className = "",
+  children = null,
+  ...props
+}: HeaderContainerProps) => {
+  return (
+    <div
+      className={`
+    juno-header-container 
+    ${!fullWidth ? "jn-w-full 2xl:jn-container 2xl:jn-mx-auto" : ""}
```
andypf, line 28: Can you confirm that this statement is correct. It looks like if not fullWidth then add `jn-w-full` class

andypf (commented):
Good Job! Please take alook at my comments

packages/ui-components/src/components/HeaderContainer/HeaderContainer.component.tsx
MartinS-git: Yes, this is correct. `jn-w-full` (default) is active on smaller displays, on larger resolutions `jn-container` is active instead

.changeset/giant-news-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudoperators/juno-ui-components": patch
+---
+
+Created HeaderContainer component to male PageHeader and TopNavigation sticky when scrolling the content. AppShell is also affected by this change
```
guoda-puidokaite, line 5: Small grammar improvement:

```suggestion
Created HeaderContainer component to make PageHeader and TopNavigation sticky when scrolling the content. AppShell is also affected by this change.
```

apps/example/src/App.jsx
```diff
@@ -59,12 +66,19 @@ const App = (props = {}) => {
       <MonorepoChecker></MonorepoChecker>
       <AsyncWorker consumerId={props.id} mockAPI={true} />
       <AppShell
+        fullWidthContent
+        embedded={props.embedded === "true" || props.embedded === true}
```
guoda-puidokaite, line 70: **(Question)**

Wondering why we check for `string` value here if `embedded` has `PropTypes.bool `?
We'll be refactoring this anyway, so can leave it. But just curious. 🙈

packages/ui-components/src/components/HeaderContainer/HeaderContainer.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import React from "react"
+
+import { HeaderContainer, HeaderContainerProps } from "./index"
+
+export default {
+  title: "Internal/HeaderContainer",
+  component: HeaderContainer,
+  argTypes: {
+    children: {
```
guoda-puidokaite, line 14: **(Blocking Improvement)**

Storybook doesn't pick up type `ReactNode` for children and reverts to `any`. Here's an [improvement](https://github.com/cloudoperators/juno/pull/601/commits/f1fa2903825034370c3449a14afaf62ae6f69686).

packages/ui-components/src/components/HeaderContainer/HeaderContainer.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+  jn-shadow-lg
+  jn-bg-theme-global-bg
+`
+
+export const HeaderContainer = ({
+  fullWidth = false,
+  className = "",
+  children = null,
+  ...props
+}: HeaderContainerProps) => {
```
guoda-puidokaite, line 23: **(Blocking Improvement)**

To improve consistency, readability and type safety 🙏: 

```suggestion
export const HeaderContainer: React.FC<HeaderContainerProps> = ({
  fullWidth = false,
  className = "",
  children = null,
  ...props
}) => {
```

packages/ui-components/src/components/HeaderContainer/HeaderContainer.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+    ${headerContainerStyles} 
+    ${className}`}
+      {...props}
+    >
+      {children}
+    </div>
+  )
+}
+
+export interface HeaderContainerProps {
```
guoda-puidokaite, line 38: **(Blocking Improvement)**

Were trying to do [this](https://github.com/cloudoperators/juno/issues/583) for all TS components:

```suggestion
export interface HeaderContainerProps extends React.HTMLAttributes<HTMLDivElement> {
```

packages/ui-components/src/components/HeaderContainer/HeaderContainer.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const HeaderContainer = ({
+  fullWidth = false,
+  className = "",
+  children = null,
+  ...props
+}: HeaderContainerProps) => {
+  return (
+    <div
+      className={`
```
guoda-puidokaite, line 26: **(Non-Blocking Improvement)**

In other components, we try extract the CSS classes into a variable so the component is easier to read.
Could we extract into `const containerStyles`, or something like that?

guoda-puidokaite (approved):
Really nice to have a component directly in TS! Super cool. 🚀

Just made some suggestions below. Feel free to only fix the blocking and just consider the rest. 💪 Thank You so much.

guoda-puidokaite (changes requested):
My brain! 🤯 I meant to request changes above.

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -55,22 +56,23 @@ export const AppShell = ({
...
-            pageHeader
-          )}
-          {topNavigation && topNavigation}
-          {/* Wrap everything except page header and footer and navigations in a main container. Add top margin to MainContainerInner as we are not in embedded mode here. */}
+          <HeaderContainer fullWidth={fullWidthContent === true} className="">
+            {pageHeader && (typeof pageHeader === "string" || pageHeader instanceof String) ? (
+              <PageHeader heading={pageHeader} />
+            ) : (
+              pageHeader
+            )}
```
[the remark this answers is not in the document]
MartinS-git: You are absolutely right with the empty className. I will remove this. I would rather not touch the other point, as I am not familiar with the background either.

apps/example/src/App.jsx
MartinS-git: Unfortunately, I can't answer that either

packages/ui-components/src/components/AppShell/AppShell.component.tsx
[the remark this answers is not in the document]
guoda-puidokaite: Thank You so much. That's no problem at all, it's not blocking. I'm looking at the components as a whole and ask to understand whether we could add some things to the improvements EPIC etc. But we should keep changes to requirements and lines touched in PR. 🙈

packages/ui-components/src/components/HeaderContainer/HeaderContainer.component.tsx
MartinS-git: I'm not quite sure about this point. Here there are already `const headerContainerStyles` with the main styles..

apps/example/src/App.jsx
andypf: The Example App was previously integrated into web pages via a <script> tag using a widget loader, which only allowed string-based props to be passed. However, we have since moved away from using the widget loader and now utilize direct dynamic imports. As a result, the reliance on string-based props is a legacy from the previous approach.

packages/ui-components/src/components/HeaderContainer/index.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export { HeaderContainer, type HeaderContainerProps } from "./HeaderContainer.component"
```
barsukov, line 6: Do we need to export it with a type?

MartinS-git: I'm not sure about that. I have looked at the other containers in the AppShell and except for `MainContainerInner`, `type` is used everywhere. In my research, I have not found any reason that speaks against it. As an expert in this topic, can you please clarify?

barsukov (changes requested):
I have a question of unnecessary check on a stirng.Maybe we could discuss it.

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -55,22 +56,23 @@ export const AppShell = ({
...
         <>
-          {pageHeader && (typeof pageHeader === "string" || pageHeader instanceof String) ? (
-            <PageHeader heading={pageHeader} />
-          ) : (
-            pageHeader
-          )}
-          {topNavigation && topNavigation}
-          {/* Wrap everything except page header and footer and navigations in a main container. Add top margin to MainContainerInner as we are not in embedded mode here. */}
+          <HeaderContainer fullWidth={fullWidthContent === true}>
+            {pageHeader && (typeof pageHeader === "string" || pageHeader instanceof String) ? (
```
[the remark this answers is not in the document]
MartinS-git: I can't tell you that either. It may be a fallback solution from earlier versions. I guess Esther or @franzheidl  have found a requirement here?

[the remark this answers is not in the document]
guoda-puidokaite: I also wondered about it here 👀 https://github.com/cloudoperators/juno/pull/607#discussion_r1843645481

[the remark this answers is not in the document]
guoda-puidokaite: Wonder if it's a sign 😅

packages/ui-components/src/components/HeaderContainer/index.ts
guoda-puidokaite: I was also wondering... now that you brought it up, @barsukov... I tried without, looks like we need to use the keyword. 🤷‍♀️
Screenshot 2024-11-19 at 11 21 51

guoda-puidokaite: But I guess this is re-exporting.

guoda-puidokaite: Should we be doing this in all components we're migrating, @barsukov? Exporting the type from the component index file.

packages/ui-components/src/components/AppShell/AppShell.component.tsx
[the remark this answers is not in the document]
guoda-puidokaite: I can do so tomorrow. But you don't have to wait for me, feel free to get a second review and dismiss my blocking request.

barsukov (approved):
LGTM

packages/ui-components/src/components/HeaderContainer/index.ts
barsukov: I think in general this concept of index files are anoying in a way I will just prefer to do export types directly, but it is ok. It looks like it is a typescript setting. So leave the type

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -36,12 +37,28 @@ export const AppShell = ({
...
+  const renderHeaderContainer = (
+    pageHeader?: AppShellProps["pageHeader"],
+    topNavigation?: AppShellProps["topNavigation"]
+  ) => {
+    if (!pageHeader && !topNavigation) {
+      return null
+    }
+    return (
+      <HeaderContainer fullWidth={fullWidthContent === true}>
+        {pageHeader && typeof pageHeader === "string" ? <PageHeader heading={pageHeader} /> : pageHeader}
```
guoda-puidokaite, line 49: **(Non-Blocking Improvement)**

We already check `if (!pageHeader && !topNavigation)`

```suggestion
        {typeof pageHeader === "string" ? <PageHeader heading={pageHeader} /> : pageHeader}
```

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -36,12 +37,28 @@ export const AppShell = ({
     )
   }
 
+  const renderHeaderContainer = (
+    pageHeader?: AppShellProps["pageHeader"],
+    topNavigation?: AppShellProps["topNavigation"]
```
guoda-puidokaite, line 42: **(Question/Non-Blocking Improvement)**

I see that this makes the type a little more reusable. 

However, we usually define types directly. One benefit is that the function is understandable to the developer without searching for the types. Makes things more readable.

Wdyt? @barsukov @MartinS-git

guoda-puidokaite (commented):
Looks great, thanks for the improvements!!! And sorry for the delay due to training. 🙏
Last few improvements. Last type related one would be the most important.

Otherwise, ready to approve. 🚀

packages/ui-components/src/components/AppShell/AppShell.component.tsx
barsukov: That was my suggestion to reuse a types, kinda best practices in this case :)

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -36,12 +37,28 @@ export const AppShell = ({
...
+    )
+  }
+
   return (
     <AppBody className={className} {...props}>
       {contentHeading || ""}
       {embedded ? (
         <>
-          {topNavigation && topNavigation}
+          {topNavigation && <HeaderContainer>{topNavigation}</HeaderContainer>}
```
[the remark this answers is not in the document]
barsukov: Not sure about it, could you double check?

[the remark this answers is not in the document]
guoda-puidokaite: Wait, nvm, oopsy. We only check both `true`. 🤯

packages/ui-components/src/components/AppShell/AppShell.component.tsx
guoda-puidokaite: Wait, nvm, oopsy. We only check if both `true`. 🤯

packages/ui-components/src/components/AppShell/AppShell.component.tsx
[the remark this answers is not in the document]
guoda-puidokaite: Should we reuse then?

```suggestion
          renderHeaderContainer(undefined, topNavigation)
```

guoda-puidokaite (approved):
Looks good! Thank You! 💪

PR #608, opened 2024-11-14
fix(greenhouse): add support_group only if supportGroups exists and is an array

ArtieReus:
# Summary

This PR improves the handling of the support_group key within initialFilters in appProps. Now, support_group is added only when authData.parsed.supportGroups exists and is a valid array. If supportGroups is undefined, null, or not an array, support_group will be omitted from initialFilters, ensuring cleaner and more reliable data handling. Fixes the issue that prevented colleagues without a supportGroups value from accessing Supernova and Doop.

Related task #611

# Changes Made

- Added a conditional spread operator to include support_group only if supportGroups is an array.
- If supportGroups does not exist or is not an array, support_group is excluded from initialFilters.

# Screenshots (if applicable)

Error when no support group is set to the logged user:

Screenshot 2024-11-14 at 12 06 54

# Testing Instructions

1. `npm i`
2. Remove from the Auth Object the supportGroup `delete authData.parsed.supportGroups`
4. `npx turbo dev --filter @cloudoperators/juno-app-greenhouse`

# Checklist

### DIFF
--- a/apps/greenhouse/src/hooks/useApi.js
+++ b/apps/greenhouse/src/hooks/useApi.js
@@ -45,11 +45,16 @@ const useApi = () => {
           const url = conf.status?.uiApplication?.url
 
           // temporary fix to forward initialFilters to the Plugins until middleware is implemented
-          const appProps = { username: authData?.parsed?.fullName }
-          appProps.initialFilters = {
-            ...(Array.isArray(authData?.parsed?.supportGroups) && {
-              support_group: authData.parsed.supportGroups.map((group) => group),
-            }),
+          // Extract username and support group information for appProps
+          const appProps = {
+            username: authData?.parsed?.fullName,
+          }
+
+          // Conditionally add initialFilters if supportGroups exists and is an array
+          if (Array.isArray(authData?.parsed?.supportGroups)) {
+            appProps.initialFilters = {
+              support_group: authData.parsed.supportGroups.map((group) => String(group)),
+            }
           }
 
           const newConf = createPluginConfig({
--- a/apps/greenhouse/src/hooks/useApi.js
+++ b/apps/greenhouse/src/hooks/useApi.js
@@ -47,7 +47,9 @@ const useApi = () => {
           // temporary fix to forward initialFilters to the Plugins until middleware is implemented
           const appProps = { username: authData?.parsed?.fullName }
           appProps.initialFilters = {
-            support_group: authData?.parsed?.supportGroups?.map((group) => group),
+            ...(Array.isArray(authData?.parsed?.supportGroups) && {
+              support_group: authData.parsed.supportGroups.map((group) => group),
+            }),
           }
 
           const newConf = createPluginConfig({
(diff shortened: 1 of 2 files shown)


apps/greenhouse/src/hooks/useApi.js
```diff
@@ -47,7 +47,9 @@ const useApi = () => {
           // temporary fix to forward initialFilters to the Plugins until middleware is implemented
           const appProps = { username: authData?.parsed?.fullName }
           appProps.initialFilters = {
-            support_group: authData?.parsed?.supportGroups?.map((group) => group),
+            ...(Array.isArray(authData?.parsed?.supportGroups) && {
```
andypf, line 50: Could you move the if statement outside of this object? If Array.isArray returns false, the entire statement will evaluate to false. I’m not sure what the result will be in that case.

andypf (commented):
Added a comment

apps/greenhouse/src/hooks/useApi.js
ArtieReus: already checked. It doesn't include any object if supportGroups doesn't exist or is not an array

ArtieReus: This is the shortest prettier way i found :)

ArtieReus: It can be tested very easy in the console:
Screenshot 2024-11-14 at 13 47 41

TilmanHaupt (dismissed):

apps/greenhouse/src/hooks/useApi.js
```diff
@@ -47,7 +47,9 @@ const useApi = () => {
...
+              support_group: authData.parsed.supportGroups.map((group) => group),
```
andypf, line 51: Please don't use map here, If authData.parsed.supportGroups is already an array, .map((group) => group) is redundant and doesn’t add value—it simply returns the same array.

TilmanHaupt (approved):

andypf (approved):
LGTM

apps/greenhouse/src/hooks/useApi.js
```diff
@@ -45,9 +45,16 @@ const useApi = () => {
...
-            support_group: authData?.parsed?.supportGroups?.map((group) => group),
+          // Extract username and support group information for appProps
+          const appProps = {
+            username: authData?.parsed?.fullName,
+          }
+
+          // Conditionally add initialFilters if supportGroups exists and is an array
+          if (Array.isArray(authData?.parsed?.supportGroups)) {
+            appProps.initialFilters = {
+              support_group: authData.parsed.supportGroups.map((group) => String(group)),
```
guoda-puidokaite, line 56: Any chance some of these could be `undefined`?

If so, suggested change would be:
`support_group: authData?.parsed?.supportGroups.map((group) => String(group)),`

ArtieReus: Thank you very much for the comment even being merged :heart:
Wouldn't be the `undefined` covered by the `if (Array.isArray(authData?.parsed?.supportGroups))`?

guoda-puidokaite: Ah, duh! 😅 Thank You!

PR #610, opened 2024-11-14
fix(supernova): activeFilters are not rendered if value is not an array

### LINKED ISSUE #609: [BUG](supernova): Non array values in FilterPills lead to failed render

TilmanHaupt:
# Summary
Fix issue where activeFilters are not rendered if the value is not an array. Closes #609

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist

- [] I have added tests that prove my fix is effective or that my feature works.
- [] I have made corresponding changes to the documentation (if applicable).

### DIFF
--- a/.changeset/clean-plums-notice.md
+++ b/.changeset/clean-plums-notice.md
@@ -1,0 +1,5 @@
+---
+"@cloudoperators/juno-app-supernova": patch
+---
+
+fix(supernova): non array values in the activeFilter are now ignored and not rendered
--- a/.changeset/clean-plums-notice.md
+++ b/.changeset/clean-plums-notice.md
@@ -1,5 +1,0 @@
----
-"@cloudoperators/juno-app-supernova": patch
----
-
-fix(supernova): non array values in the activeFilter are now ignored and not rendered
--- a/apps/supernova/src/components/filters/FilterPills.jsx
+++ b/apps/supernova/src/components/filters/FilterPills.jsx
@@ -30,28 +30,30 @@ const FilterPills = () => {
   return (
     <Stack gap="2" wrap={true}>
       {Object.entries(activeFilters).map(([key, filterItems]) => {
-        return filterItems.map((item) =>
-          pausedFilters[key]?.includes(item) ? (
-            <Pill
-              className="bg-theme-background-lvl-4 opacity-70	"
-              pillKey={key}
-              pillValue={item}
-              closeable
-              onClose={() => deleteFilter(key, item)}
-              key={`${key}:${item}`}
-              onClick={() => activateFilter(key, item)}
-            />
-          ) : (
-            <Pill
-              pillKey={key}
-              pillValue={item}
-              closeable
-              onClose={() => deleteFilter(key, item)}
-              key={`${key}:${item}`}
-              onClick={() => pauseFilter(key, item)}
-            />
+        if (Array.isArray(filterItems)) {
+          return filterItems.map((item) =>
+            pausedFilters[key]?.includes(item) ? (
+              <Pill
+                className="bg-theme-background-lvl-4 opacity-70	"
+                pillKey={key}
+                pillValue={item}
+                closeable
+                onClose={() => deleteFilter(key, item)}
+                key={`${key}:${item}`}
+                onClick={() => activateFilter(key, item)}
+              />
+            ) : (
+              <Pill
+                pillKey={key}
+                pillValue={item}
+                closeable
+                onClose={() => deleteFilter(key, item)}
+                key={`${key}:${item}`}
+                onClick={() => pauseFilter(key, item)}
+              />
+            )
           )
-        )
+        }
       })}
     </Stack>
   )
--- a/apps/supernova/src/components/filters/FilterPills.jsx
+++ b/apps/supernova/src/components/filters/FilterPills.jsx
@@ -31,7 +31,7 @@ const FilterPills = () => {
     <Stack gap="2" wrap={true}>
       {Object.entries(activeFilters).map(([key, filterItems]) => {
         if (Array.isArray(filterItems)) {
-          return filterItems.map((item) =>
+          return filterItems?.map((item) =>
             pausedFilters[key]?.includes(item) ? (
               <Pill
                 className="bg-theme-background-lvl-4 opacity-70	"
--- a/apps/supernova/src/components/filters/FilterPills.jsx
+++ b/apps/supernova/src/components/filters/FilterPills.jsx
@@ -30,30 +30,28 @@ const FilterPills = () => {
   return (
     <Stack gap="2" wrap={true}>
       {Object.entries(activeFilters).map(([key, filterItems]) => {
-        if (Array.isArray(filterItems)) {
-          return filterItems?.map((item) =>
-            pausedFilters[key]?.includes(item) ? (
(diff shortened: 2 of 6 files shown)


apps/supernova/src/components/filters/FilterPills.jsx
```diff
@@ -30,28 +30,30 @@ const FilterPills = () => {
...
-            <Pill
-              pillKey={key}
-              pillValue={item}
-              closeable
-              onClose={() => deleteFilter(key, item)}
-              key={`${key}:${item}`}
-              onClick={() => pauseFilter(key, item)}
-            />
+        if (Array.isArray(filterItems)) {
+          return filterItems.map((item) =>
```
guoda-puidokaite, line 34: Ah! `Map` was being called directly, assuming an array. 🙈

Slightly cleaner way to do this:
```suggestion
          return filterItems?.map((item) =>
```

guoda-puidokaite (commented):
Nice catch, just slight suggestion.

guoda-puidokaite (changes requested):
Could we add a unit test to cover this bug for the future?

guoda-puidokaite:
> Could we add a unit test to cover this bug for the future?

Sorry, already done

apps/supernova/src/components/filters/FilterPills.jsx
```diff
@@ -30,28 +30,30 @@ const FilterPills = () => {
...
+            pausedFilters[key]?.includes(item) ? (
+              <Pill
+                className="bg-theme-background-lvl-4 opacity-70	"
```
guoda-puidokaite, line 37: Tiny 😅

```suggestion
                className="bg-theme-background-lvl-4 opacity-70"
```

apps/supernova/src/components/filters/FilterPills.jsx
```diff
@@ -30,28 +30,30 @@ const FilterPills = () => {
...
-          ) : (
-            <Pill
-              pillKey={key}
-              pillValue={item}
-              closeable
-              onClose={() => deleteFilter(key, item)}
-              key={`${key}:${item}`}
-              onClick={() => pauseFilter(key, item)}
-            />
+        if (Array.isArray(filterItems)) {
```
guoda-puidokaite, line 33: This line would then be unnecessary with optional mapping

```suggestion
        if (Array.isArray(filterItems)) {
```

guoda-puidokaite: Up to you which way you prefer. I just like the consistency and shortness of the optional solution.

apps/supernova/src/components/filters/FilterPills.jsx
ArtieReus: I like this way looks cleaner

.changeset/clean-plums-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudoperators/juno-app-supernova": patch
```
hodanoori, line 2: Shouldn't you add a patch changeset also for greenhouse?

ArtieReus (approved):
lgtm

hodanoori (dismissed):
LGTM

ArtieReus (approved):
lgtm

apps/supernova/src/lib/createFiltersSlice.jsx
```diff
@@ -28,11 +28,21 @@ const parsePredefinedFilters = (predefinedFilters) => {
...
+    if (Array.isArray(value)) {
+      acc[key] = value // valid key-value pair
+    } else {
+      console.warn(`[supernova]::parseInitialFilters: Value for "${key}" is not an Array.`)
+    }
+    return acc
+  }, {})
+
   // Check if all keys are in filterLabelValues
   if (!Object.keys(initialFilters).every((key) => filterLabels.includes(key))) {
```
guoda-puidokaite, line 47: **(Non-blocking concerns)**

Concerned that there's other string/array methods in this file that don't check valid type before executing e.g. `.included`.
Your changes look good, but if we've the time (even in a follow-up) PR would be great to ensure that we prevent the same issue causing bugs, at least in the same files.

apps/supernova/src/lib/createFiltersSlice.jsx
```diff
@@ -28,11 +28,21 @@ const parsePredefinedFilters = (predefinedFilters) => {
 const parseInitialFilters = (initialFilters, filterLabels) => {
   if (!initialFilters) return initialFiltersState.initialFilters
 
-  if (typeof initialFilters !== "object") {
+  if (typeof initialFilters !== "object" || initialFilters === null) {
```
guoda-puidokaite, line 31: **(Non-blocking suggestion)**

Safer coverage?

```suggestion
  if (!initialFilters || typeof initialFilters !== "object")
```

TilmanHaupt: its not safer because undefined is not an object.

ArtieReus: cool @guoda-puidokaite 

!initialFilters checks if initialFilters is falsy (i.e., null, undefined, false, 0, NaN, or an empty string).

guoda-puidokaite: > its not safer because undefined is not an object.

Not sure what you mean by this? I just think additionally to checking for `null`, we could cover all `falsy` values. But maybe not necessary here. You know code more @TilmanHaupt @ArtieReus

apps/supernova/src/lib/createFiltersSlice.jsx
TilmanHaupt: I see none. At least on the surface everything is tested. Maybe there are like here this problems in nested forms.

apps/supernova/src/lib/createFiltersSlice.jsx
```diff
@@ -28,11 +28,21 @@ const parsePredefinedFilters = (predefinedFilters) => {
...
     console.warn("[supernova]::parseInitialFilters: initialFilters object is not an object")
     return {}
   }
 
+  // Check if all values are arrays
+  initialFilters = Object.entries(initialFilters).reduce((acc, [key, value]) => {
```
guoda-puidokaite, line 37: **(Non-blocking suggestion)**

Good solution, wonder if this would be more readable instead of manually constructing key-value pairs. But please merge as you want as we need to do this fast, no problem.

```suggestion
  initialFilters = Object.fromEntries(
    Object.entries(initialFilters).filter(([key, value]) => {
      const isArray = Array.isArray(value)
      if (!isArray) {
        console.warn(`[supernova]::parseInitialFilters: Value for "${key}" is not an Array.`)
      }
      return isArray
    })
  )
```

apps/supernova/src/lib/createFiltersSlice.jsx
TilmanHaupt: All other values are not objects.

andypf (approved):
👍

apps/supernova/src/lib/createFiltersSlice.jsx
guoda-puidokaite: Ahhh, duh. Nice.

PR #615, opened 2024-11-14
chore(supernova): adjust Dockerfile to address vite build

andypf:
# Summary

Adjusted the Dockerfile to align with the new Vite-based build system for the Supernova app.

# Changes Made

- Adjusted the Dockerfile to align with the new Vite-based build system for the Supernova app.

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


apps/supernova/docker/Dockerfile
```diff
@@ -7,64 +7,17 @@ LABEL org.opencontainers.image.source=https://github.com/cloudoperators/juno/tre
...
+RUN apk add --no-cache jq && apk upgrade --no-cache --no-progress 
 
-RUN apk add --no-cache jq \
-  && apk upgrade --no-cache --no-progress \
-  && apk del --no-cache --no-progress apk-tools alpine-keys
-
-COPY --from=build /app /usr/share/nginx/html
+WORKDIR /usr/share/nginx/html
 
 ENV THEME=""
```
ArtieReus, line 22: Can we automate the creation of environment variables based on the `appProps.template.json` file? As mentioned earlier, we could use `appProps.template.json` to generate `appProps.json`, and it could also be leveraged for this step.

Something like this:
```
# Loop through the JSON keys and create ENV variables
for key in $(jq -r 'keys_unsorted[]' "$json_file"); do
  value=$(jq -r ".\"$key\"" "$json_file")
  echo "ENV $key=\"$value\"" >> "$env_file"
done
```

apps/supernova/docker/entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+update_json_field() {
+  field="$1"
+  value="$2"
+  if [ -n "$value" ]; then
+    # Ensure jq and temporary file handling are POSIX compliant
+    jq --arg v "$value" ".${field} = \$v" "$APP_PROPS_FILE" >tmp.json && mv tmp.json "$APP_PROPS_FILE"
+  fi
+}
+
+# Update fields if corresponding environment variables are set
```
ArtieReus, line 24: can we also automate this reading the `appProps.template.json` and mapping the env variable? when adding new attributes we then have con change:
- appProps.template.json
- Readme.md
- Dockerfile
- entrypoint.sh

apps/supernova/docker/entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+update_json_field "filterLabels" "$FILTER_LABELS"
+update_json_field "silenceExcludedLabels" "$SILENCE_EXCLUDED_LABELS"
+update_json_field "silenceTemplates" "$SILENCE_TEMPLATES"
+update_json_field "predefinedFilters" "$PREDEFINED_FILTERS"
+
+echo "$APP_PROPS_FILE has been updated."
+
+cat "$APP_PROPS_FILE"
+
+/docker-entrypoint.sh
```
ArtieReus, line 36: how is this script generated?

apps/supernova/docker/Dockerfile
```diff
@@ -73,56 +26,25 @@ ENV SILENCE_EXCLUDED_LABELS=""
...
-# Function to update JSON fields using jq
-update_json_field() {
-  field="\$1"
-  value="\$2"
-  if [ -n "\$value" ]; then
-    # Ensure jq and temporary file handling are POSIX compliant
-    jq --arg v "\$value" ".\${field} = \\\$v" "\$APP_PROPS_FILE" > tmp.json && mv tmp.json "\$APP_PROPS_FILE"
-  fi
+# default appProps.json
+RUN cat <<EOF > /appProps.json
```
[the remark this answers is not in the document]
andypf: That could be a good idea. I suggest creating a task with the proposed improvements, so they can be implemented across all apps at once. However, this PR focuses specifically on adjusting the Dockerfile to ensure the image can be built again.

apps/supernova/docker/entrypoint.sh
andypf: it is from nginx

ArtieReus (approved):
lgtm But would you please create a task for the Improvements @andypf ?

hgw77 (approved):

apps/supernova/docker/Dockerfile
```diff
@@ -73,56 +26,25 @@ ENV SILENCE_EXCLUDED_LABELS=""
...
-cat "\$APP_PROPS_FILE"
+COPY --from=build /app/apps/supernova/build /usr/share/nginx/html/
 
-exec "\$@"
-EOF
+COPY apps/greenhouse/docker/nginx.conf /etc/nginx/conf.d/default.conf
+COPY apps/greenhouse/docker/entrypoint.sh /usr/local/bin/entrypoint.sh 
 
-RUN chmod +x /supernova_entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
```
hgw77, line 45: this step could be removed when we change the file permissions before checking in. Just do `chmod +x /usr/local/bin/entrypoint.sh` and push the changes.

On the other side you need the `entrypoint.sh` as executable to run it for test reasons on your dev system so it is always a good idea to made it executable by default :-)

andypf: Good catch! Let’s make sure to add it to the list of improvements.

PR #640, opened 2024-11-26
chore(ui): migrate DateTimePicker to TypeScript

gjaskiewicz:
# Summary

migrate DateTimePicker to TypeScript

# Changes Made

- migrate DateTimePicker to TypeScript

# Related Issues

- Issue 1: [[link to issue](https://github.com/cloudoperators/juno/issues/227)]

# Testing Instructions

1. `npm i`
2. `npm run storybook`

# Checklist


packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,640 @@
...
+      )}
+      {helptext && hasLength(helptext) ? <FormHint text={helptext} className="jn-mt-0" /> : ""}
+    </div>
+  )
+}
+
+// eslint-disable-next-line no-unused-vars
+type DateChangeHandler = (date: Date[] | undefined, dateStr: string | undefined) => void
+
+export interface DateTimePickerProps {
```
guoda-puidokaite, line 543: **(Blocking)**

Please extend with intrinsic props

guoda-puidokaite (changes requested):
Nice job! So many lines in the component. 🚀
I know this PR is still a draft but I already reviewed to give early feedback... Added a few blocking improvements.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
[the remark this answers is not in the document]
guoda-puidokaite: Example (might need to exclude some props to avoid conflicts):

```suggestion
export interface DateTimePickerProps extends React.HTMLAttributes<HTMLInputElement> {
```

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,640 @@
...
+  width = "full",
+  wrapperClassName = "",
+  ...props
+}) => {
+  // always generate auto-id string using the useId hook to avoid "more hooks than in previous render" error when removing custom id:
+  const autoId = "juno-datetimepicker-" + useId()
+  const theId = id && id.length ? id : autoId
+
+  const fpRef = useRef<HTMLInputElement>(null) // the dom node flatpickr instance will be bound to
+  const flatpickrInstanceRef = useRef<flatpickr.Instance | {}>({}) // The actual flatpickr instance
```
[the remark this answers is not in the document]
gjaskiewicz: After talking f2f, I decided to give it a try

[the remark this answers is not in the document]
gjaskiewicz: ok, seems like to be working

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,640 @@
...
+  /** Whether the DateTimePicker selected date was negatively validated.  */
+  invalid?: boolean
+  /** The label of the DateTimePicker input element. */
+  label?: string
+  /** Localization string or object. Can be used to set starting day of the week, e.g. Mondays instead of Sundays. More on localization: https://flatpickr.js.org/localization/ */
+  locale?: LocaleKey | Partial<CustomLocale> | null
+  /** The maximum / latest date a user can select (inclusive). */
+  maxDate?: DateOption | null
+  /** The minimum / earliest date a user can select (inclusive). */
+  minDate?: DateOption | null
```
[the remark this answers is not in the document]
gjaskiewicz: Resolved all 'any' in storybook

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,640 @@
...
+  /** The number of months to show in the date picker */
+  showMonths?: number
+  /** A text to render when the DateTimePicker was successfully validated */
+  successtext?: React.ReactNode
+  /** Displays time picker in 24 hour mode without AM/PM selection when enabled. Requires `enableTime` to be set, too. Default is `false`. */
+  time_24hr?: boolean
+  /** Whether the DateTimePicker has been successfully validated */
+  valid?: boolean
+  /** The value of the datepicker. Date Objects, timestamps, ISO date strings, chronological date strings `YYYY-MM-DD HH:MM` (must be compatible to current `dateFormat`), and the shortcut `today` are all accepted. */
+  value?: DateOption | DateOption[] | null
```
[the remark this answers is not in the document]
gjaskiewicz: Resolved all 'any' in storybook

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,640 @@
...
+  /** How the `aria-label` date for each day in the calendar will be formed. Uses the same rules/tokens as `dateFormat´ as described here: https://flatpickr.js.org/formatting/. When changing this, make sure the outcome makes sense when using a screenreader.*/
+  ariaDateFormat?: string
+  /** Pass custom classNames. These will be appended to the input element of the DateTimePicker. */
+  className?: string
+  /** A custom string to separate individual dates in `multiple` mode. */
+  conjunction?: string
+  /** A string of characters to customize how a date will be formatted in the input field. Available options: https://flatpickr.js.org/formatting/ */
+  dateFormat?: string
+  /** Sets the default date of the DateTimePicker. Same as `value`, only here for compatibility with the original Flatpickr library. If both `value` and `defaultDate` are being passed, `value` will win. Date Objects, timestamps, ISO date strings, chronological date strings `YYYY-MM-DD HH:MM` (must be compatible to current `dateFormat`), and the shortcut `today` are all accepted. */
+  defaultDate?: DateOption | DateOption[] | null
```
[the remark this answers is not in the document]
gjaskiewicz: Resolved all 'any' in storybook

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,640 @@
...
+  /** The initial value of the hour input element. Only effective if time is enabled. Note this will only set the hour input element to the value specified. Setting this options will not set a selected value on the DateTimePicker. */
+  defaultHour?: number
+  /** The initial value of the minute input element. Only effective if time is enabled. Note this will only set the minute input element to the value specified. Setting this options will not set a selected value on the DateTimePicker. */
+  defaultMinute?: number
+  /** Same as value, defaultDate */
+  defaultValue?: DateOption | DateOption[]
```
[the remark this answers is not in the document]
gjaskiewicz: Resolved all 'any' in storybook

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,640 @@
...
+  /** The mode of the Datepicker. */
+  mode?: "single" | "multiple" | "range"
+  /** Whether to show a dropdown to select the current month. Default is "static". If `showMonths` is set to be greater than 1, it will always be displayed as static. Arrows to scroll through the months as well as through years will still be displayed and working. */
+  monthSelectorType?: "static" | "dropdown"
+  /** The name of the DateTimePicker input element */
+  name?: string
+  /** Set to `true` to not display a calendar at all. To create a time picker, set `enableTime` to true, too. */
+  noCalendar?: boolean
+  /** A handler to be executed when the DateTimePicker input element looses focus. */
+  onBlur?: DateChangeHandler
```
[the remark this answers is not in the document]
gjaskiewicz: Done

guoda-puidokaite (changes requested):
Thank You for the changes! 🚀 Just a few blocking changes again.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,651 @@
...
+  /** Whether the DateTimePicker selected date was negatively validated.  */
+  invalid?: boolean
+  /** The label of the DateTimePicker input element. */
+  label?: string
+  /** Localization string or object. Can be used to set starting day of the week, e.g. Mondays instead of Sundays. More on localization: https://flatpickr.js.org/localization/ */
+  locale?: LocaleKey | Partial<CustomLocale> | null
+  /** The maximum / latest date a user can select (inclusive). */
+  maxDate?: DateOption | null
+  /** The minimum / earliest date a user can select (inclusive). */
+  minDate?: DateOption | null
```
guoda-puidokaite, line 591: **(Breaking Change)**

`DateOption` is the most accurate type here. But we need to be backwards compatible as we allowed `string | array | object | number` previously. We can remove these in the future (added to improvements).

Check all other types for backward compatibility too, bitte. I see there's a few others. 💪🤪

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,651 @@
...
+  /** How the `aria-label` date for each day in the calendar will be formed. Uses the same rules/tokens as `dateFormat´ as described here: https://flatpickr.js.org/formatting/. When changing this, make sure the outcome makes sense when using a screenreader.*/
+  ariaDateFormat?: string
+  /** Pass custom classNames. These will be appended to the input element of the DateTimePicker. */
+  className?: string
+  /** A custom string to separate individual dates in `multiple` mode. */
+  conjunction?: string
+  /** A string of characters to customize how a date will be formatted in the input field. Available options: https://flatpickr.js.org/formatting/ */
+  dateFormat?: string
+  /** Sets the default date of the DateTimePicker. Same as `value`, only here for compatibility with the original Flatpickr library. If both `value` and `defaultDate` are being passed, `value` will win. Date Objects, timestamps, ISO date strings, chronological date strings `YYYY-MM-DD HH:MM` (must be compatible to current `dateFormat`), and the shortcut `today` are all accepted. */
+  defaultDate?: DateOption | DateOption[] | null
```
guoda-puidokaite, line 559: **(Blocking Change)**

We don't need to explicitly allow `null` in these props, I think? It's optional, but we didn't support explicit `null` in the past.

Please double check with for all other props too. I see at least five. Removing one causes Storybook file errors, we need to check what was supported in the past etc.

guoda-puidokaite (changes requested):
Requesting changes for above

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,651 @@
...
+  onMonthChange?: (date?: Date[], dateStr?: string) => void
+  /** A handler to be executed when the DateTimePicker calendar opens */
+  // eslint-disable-next-line no-unused-vars
+  onOpen?: (date?: Date[], dateStr?: string) => void
+  /** A handler to be executed when the DateTimePicker component is ready */
+  // eslint-disable-next-line no-unused-vars
+  onReady?: (date?: Date[], dateStr?: string) => void
+  /** A handler to be executed when the selected year changes */
+  // eslint-disable-next-line no-unused-vars
+  onYearChange?: (date?: Date[], dateStr?: string) => void
```
guoda-puidokaite, line 628: **(Blocking)**

Please make sure these method inputs etc. are backwards compatible too.

guoda-puidokaite (changes requested):
Let's double check backward compatibility for all props

andypf (dismissed):
LGTM! However, please make sure to address the issues identified by Guoda.

guoda-puidokaite:
Side note: Please add issue links to your PR, following the template. 🙏

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: I check for nulls and made updates

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
guoda-puidokaite: **Update:** I added `DateOption` as an [improvement](https://github.com/cloudoperators/juno/issues/570) for date props in this component. For now, let's just support types that were previously supported and leave it out.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,651 @@
...
+  /** The number of months to show in the date picker */
+  showMonths?: number
+  /** A text to render when the DateTimePicker was successfully validated */
+  successtext?: React.ReactNode
+  /** Displays time picker in 24 hour mode without AM/PM selection when enabled. Requires `enableTime` to be set, too. Default is `false`. */
+  time_24hr?: boolean
+  /** Whether the DateTimePicker has been successfully validated */
+  valid?: boolean
+  /** The value of the datepicker. Date Objects, timestamps, ISO date strings, chronological date strings `YYYY-MM-DD HH:MM` (must be compatible to current `dateFormat`), and the shortcut `today` are all accepted. */
+  value?: DateOption | DateOption[]
```
guoda-puidokaite, line 644: **(Breaking Change)**

Please also have a look at this

guoda-puidokaite (changes requested):
🙏

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
andypf: In this case, the functions are custom implementations and are unrelated to HTMLElement. While they share the same name, their functionality differs. Grzegorz implemented them correctly, as evidenced by how they are being used in the code.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
andypf: Since maxDate and minDate are not being manipulated and are simply passed through to the flatpickr, the DateOption type is appropriate for them.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: I made it match "func" as closesly as possible in recent update

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: Updated type

andypf: Here, value should accept the types string | number | object | array. This is necessary because the value is transformed into a string within the component:
```js
const stringifiedValue = JSON.stringify(value)
```

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
guoda-puidokaite: Thanks, @andypf. The current implementation was [my suggestion](https://github.com/cloudoperators/juno/pull/640#discussion_r1860374989). Just when I was thinking about it again, I wanted to make sure that it's backward compatible.

In [this comment](https://github.com/cloudoperators/juno/pull/640#discussion_r1863145788) I was just sharing examples where the input was more general. In this case, it could be e.g. `(event?: any) => void`.

I'm just trying to make sure my suggestion was correct for what was allowed in the past.

And `(date?: Date[], dateStr?: string) => void` would be an improvement. Similar to backward compatibility consideration `[here](https://github.com/cloudoperators/juno/pull/640#discussion_r1863406393)`.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,651 @@
...
+      defaultDate: (defaultDate || defaultValue) as DateOption | DateOption[],
+      defaultHour: defaultHour,
+      defaultMinute: defaultMinute,
+      disable: disable as DateLimit<DateOption>[],
+      enableSeconds: enableSeconds,
+      enableTime: enableTime,
+      hourIncrement: hourIncrement,
+      locale: (locale || undefined) as LocaleKey | Partial<CustomLocale>,
+      maxDate: (maxDate || undefined) as DateOption,
+      minDate: (minDate || undefined) as DateOption,
```
guoda-puidokaite, line 273: Converting/checking for a valid date type would be nicer as there's potential to cast incorrect values to a date type and then treat it as such, without any checks.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,651 @@
...
+  /** How the `aria-label` date for each day in the calendar will be formed. Uses the same rules/tokens as `dateFormat´ as described here: https://flatpickr.js.org/formatting/. When changing this, make sure the outcome makes sense when using a screenreader.*/
+  ariaDateFormat?: string
+  /** Pass custom classNames. These will be appended to the input element of the DateTimePicker. */
+  className?: string
+  /** A custom string to separate individual dates in `multiple` mode. */
+  conjunction?: string
+  /** A string of characters to customize how a date will be formatted in the input field. Available options: https://flatpickr.js.org/formatting/ */
+  dateFormat?: string
+  /** Sets the default date of the DateTimePicker. Same as `value`, only here for compatibility with the original Flatpickr library. If both `value` and `defaultDate` are being passed, `value` will win. Date Objects, timestamps, ISO date strings, chronological date strings `YYYY-MM-DD HH:MM` (must be compatible to current `dateFormat`), and the shortcut `today` are all accepted. */
+  defaultDate?: string | any[] | object | number
```
guoda-puidokaite, line 559: **Non-Blocking Suggestion**

Create a reusable `date` type?

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: explained in f2f conversation that it is ok in this case

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: done

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: It has been adressed

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,700 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useEffect, useRef, useId, useMemo, useState } from "react"
+import flatpickr from "flatpickr"
+import { FormHint } from "../FormHint/"
+import { Icon } from "../Icon"
+import { Label } from "../Label"
+import "./datetimepicker.scss"
```
barsukov, line 11: **(Blocking)**
Please move it down in the end of imports like in original file

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,700 @@
...
+  // variables starting with underscore are not linted
+  const [_hasFocus, setHasFocus] = useState<boolean>(false)
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const [isInvalid, setIsInvalid] = useState<boolean>(false)
+  const [isValid, setIsValid] = useState<boolean>(false)
+
+  const hasLength = (node: React.ReactNode) => (typeof node === "string" || Array.isArray(node)) && node.length
+
+  const updateFlatpickrInstance = (newKeys: flatpickr.Instance | null) => {
+    if (flatpickrInstanceRef.current === null) {
```
barsukov, line 159: **(Question)**

That could be a potential breaking change or fix since you change the logic that was before in this method to use object.keys, please provide explanation for that.

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
```diff
@@ -0,0 +1,700 @@
...
+      {helptext && hasLength(helptext) ? <FormHint text={helptext} className="jn-mt-0" /> : ""}
+    </div>
+  )
+}
+
+// eslint-disable-next-line no-unused-vars
+type DateChangeHandler = (dates?: Date[], dateStr?: string, instance?: flatpickr.Instance) => void
+
+// eslint-disable-next-line no-unused-vars
+type LegacyEventHandler = (...args: any) => any
```
barsukov, line 548: use unknown isntead of any

barsukov (changes requested):
I left few comments please answer them and fix some of them the main issue is with this method `updateFlatpickrInstance` you modified which could produce a potential bugs or fix :)

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: It won't allow for backward compatibility. Previously this was arbitrary function, I think 'any' closely matches such use case while 'unknown' doesn't. eg.

const OnYearChange = (_a :string) :void => { }

this gets accepted with 'any' (pre-migration case) and does not with 'unknown'

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: I had some serious troubles solution with object.keys to work in typescript, while it seemed to just do what current script does.
Could you describe the difference, so that I could know and try to make correction?

packages/ui-components/src/components/DateTimePicker/DateTimePicker.test.tsx
```diff
@@ -0,0 +1,977 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from "react"
+import { cleanup, render, screen, waitFor } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { DateTimePicker } from "./index"
+import { AppShellProvider } from "../AppShellProvider/AppShellProvider.component"
```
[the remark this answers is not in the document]
gjaskiewicz: Portal works here. Changed

guoda-puidokaite (changes requested):
**(Regression? Please double check...)**

In the PR Storybook (right), it seems to take two clicks to set or change a date compared to live (left). 
Maybe something changed regarding state setting or so? 
Could you double check 1. whether you can recreate this 2. why it's happening and fix it. @gjaskiewicz 

https://github.com/user-attachments/assets/0ef74ed0-f50c-4343-82a0-6e8ac84ae922

guoda-puidokaite (commented):
**(Unrequested Change)**

Default has been changed. Looks fine, but can we keep everything the same?

guoda-puidokaite:
**(Requested Improvement)**

For all deprecated props, could you mention something along the lines of `Some prop types have been deprecated: Use x and y for future compatibility`. As this current message is usually for a deprecated component.

Screenshot 2024-12-04 at 10 00 14
Screenshot 2024-12-04 at 10 01 11

packages/ui-components/src/components/DateTimePicker/DateTimePicker.component.tsx
gjaskiewicz: I left the new one, because I don't see functional difference. However there is also this equivalent

  const updateFlatpickrInstance = (newKeys: flatpickr.Instance | null) =>
    (flatpickrInstanceRef.current = {
      ...flatpickrInstanceRef.current,
      ...Object.keys(newKeys!).reduce((a, key) => {
        const k = key as keyof flatpickr.Instance
        const newVal = newKeys![key as keyof flatpickr.Instance]
        a[k] = newVal as never
        return a
      }, {} as flatpickr.Instance),
    } as flatpickr.Instance)
    
    what I'm unhappy about here is amount of casting and casting to 'never' - not sure if this would even pass verification pipeline.

guoda-puidokaite (approved):
Regression fixed, Thank You. 
Types look good. I've added types for this component to future improvements. Can be re-addressed there.

barsukov (approved):

PR #673, opened 2024-12-10
chore(ui): integrate vite for ui-components

barsukov:
# Summary

This pull request transitions the project from Rollup to Vite as the primary bundler to improve build performance and developer experience. It maintains backward compatibility while ensuring existing workflows remain functional. The Storybook configuration remains untouched for now, with plans to update it in a subsequent iteration.

# Changes Made

- Removed Rollup as the bundler.
- Fully integrated Vite as the new bundler across the project.
- Ensured backward compatibility with existing configurations and workflows.

# Related Issues

- Issue: https://github.com/cloudoperators/juno/issues/444

# Screenshots (if applicable)

# Testing Instructions

1. Run `npm install` to install updated dependencies.
2. Execute `npm run build` to verify the Vite setup.
3. Test backward compatibility with previous configurations.

# Checklist

- [ ] Performed a self-review of the changes.
- [ ] Verified backward compatibility with existing setups.
- [ ] Added comments for any complex sections of code.
- [ ] Updated or ensured no changes were required for documentation.
- [ ] Tested locally, and all unit tests pass.
- [ ] Ensured no new warnings or errors are introduced.

This marks the initial step toward optimizing our build process with Vite while keeping Storybook configurations pending for a future update.


apps/carbon/src/App.tsx
```diff
@@ -36,7 +36,7 @@ const extensions = [
 const App = (props: AppProps) => (
   <ErrorBoundary>
     <AppShellProvider theme={`${props.theme ? props.theme : "theme-dark"}`}>
-      <AppShell embedded>
+      <AppShell embedded fullWidthContent={undefined} sideNavigation={undefined} topNavigation={undefined}>
```
taymoor89, line 39: do we really need to supply undefined for optional props?

andypf, line 39: This looks strange to me. I suspect it might be related to the type definitions in the ui-components. Shouldn’t those types be optional?

apps/carbon/src/components/AppContent/NoExtensionFound.tsx
```diff
@@ -8,7 +8,9 @@ import { Message } from "@cloudoperators/juno-ui-components"
 
 const NoExtensionFound = () => (
   <div className="p-4">
```
andypf, line 10: same :(

apps/template/src/App.tsx
```diff
@@ -12,7 +12,7 @@ import { ErrorBoundary } from "react-error-boundary"
 interface AppProps {
   theme?: string
   embedded?: string | boolean
-  endpoint?: string
+  endpoint?: string | null | undefined
```
taymoor89, line 15: I think there is no need for an explicit `undefined` because `?` makes a property optional which already covers that case.

apps/template/src/App.tsx
```diff
@@ -39,20 +39,17 @@ export const App = (props: AppProps) => {
...
   }
 
   return (
     <QueryClientProvider client={queryClient}>
       <AppShell
         pageHeader="Converged Cloud | App Template"
         embedded={props.embedded === "true" || props.embedded === true}
+        fullWidthContent={false}
+        sideNavigation={null}
+        topNavigation={null}
```
taymoor89, line 52: aren't these props optional why do we need to supply falsy values?

apps/carbon/src/App.tsx
barsukov: Totally, right I did it only because when we start exporting types the app started failed, so I thought it would be a part of diferent PR to normalize types and havea. ticket for that.

apps/carbon/src/components/AppContent/NoExtensionFound.tsx
barsukov: Same

apps/carbon/src/App.tsx
barsukov: No, we do not but during the migration it was a mistake, so now we have a lot of left overs that we need to fix. This is a temporary solution. Plan is to move on and fix all leftovers

apps/template/src/App.tsx
barsukov: agree!

apps/template/src/App.tsx
barsukov: For same weird reasons I will have a look how I can configure it better

barsukov:
@taymoor89 I create an extra ticket for the lefotver typechek, So i will resolve the comments because as you can see the PR is already very big and I think we can do follow ups on this one.

apps/carbon/src/App.tsx
taymoor89: agree, let's move on.

taymoor89 (approved):

andypf (dismissed):

guoda-puidokaite (changes requested):
Reviewing. I'd like to understand some of the necessary changes in the PR.

apps/template/src/App.tsx
```diff
@@ -39,20 +39,17 @@ export const App = (props: AppProps) => {
...
-    )
+    return <div className="w-1/2">{error?.message || error?.toString() || "An error occurred"}</div>
   }
 
   return (
     <QueryClientProvider client={queryClient}>
       <AppShell
         pageHeader="Converged Cloud | App Template"
         embedded={props.embedded === "true" || props.embedded === true}
+        fullWidthContent={false}
```
guoda-puidokaite, line 50: **(Change Required ?)**

When `embedded` is `true`, default `fullWidthContent` in `AppShell` is `true`. And when not, `false`. 
Is this change just to fix the TS issues? If so, since it's not part of requirements, let's just assign `undefined`, to ensure we're not changing any functionality? Or are we sure it should be `false`?

apps/template/src/components/AppContent.tsx
```diff
@@ -8,11 +8,11 @@ import { Breadcrumb, BreadcrumbItem, Container } from "@cloudoperators/juno-ui-c
 
 import WelcomeView from "./WelcomeView"
 
-const AppContent = () => {
+const AppContent = (): JSX.Element => {
   return (
     <>
       <Breadcrumb>
-        <BreadcrumbItem icon="home" label="Template App Home" />
+        <BreadcrumbItem onClick={() => {}} icon={null} label="Template App Home" />
```
guoda-puidokaite, line 15: **(Curious Question ?)**

Why did we change `icon` to `null` here? 
- The default icon seems to be `help`, so I think this would change styling? Wdyt?
- `icon` is already `null` in the component, by default. I don't think we need to explicitly set it here? If TS is complaining, then this is an interesting case where a default set doesn't fix the issue. 😳

guoda-puidokaite (commented):
PR looks good overall, Thank You. Just have a few questions on changes not in requirements or explanation, would prefer to leave those lines untouched... unless we're sure? 🙏🚀💪

apps/template/src/App.tsx
```diff
@@ -39,20 +39,17 @@ export const App = (props: AppProps) => {
...
 
   const fallbackRender = ({ error }: { error: { message: string } }) => {
-    return (
-      <div className="w-1/2">
-        <CodeBlock className={preErrorClasses} copy={false}>
-          {error?.message || error?.toString() || "An error occurred"}
-        </CodeBlock>
-      </div>
-    )
+    return <div className="w-1/2">{error?.message || error?.toString() || "An error occurred"}</div>
```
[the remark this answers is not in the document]
barsukov: I just could not figure out how to fix in this case I saw since this app does not really proper in productive I think it would be makes more sense to revert this changes and just ts-ignore them for now until we figure it out fully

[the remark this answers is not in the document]
barsukov: WDYT?

apps/template/src/App.tsx
```diff
@@ -46,13 +46,17 @@ export const App = (props: AppProps) => {
         </CodeBlock>
       </div>
     )
+    // return <div className="w-1/2">{error?.message || error?.toString() || "An error occurred"}</div>
```
guoda-puidokaite, line 49: Can we remove commented code if all fine?

apps/template/src/App.tsx
guoda-puidokaite: Still same, can you try revert? 🙏

apps/template/src/App.tsx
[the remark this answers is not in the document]
guoda-puidokaite: Looks good reverted, thanks.

apps/template/src/components/AppContent.tsx
guoda-puidokaite: What about this?

apps/template/src/App.tsx
guoda-puidokaite: Still same, can you try revert? 🙏

apps/template/src/components/AppContent.tsx
```diff
@@ -8,11 +8,11 @@ import { Breadcrumb, BreadcrumbItem, Container } from "@cloudoperators/juno-ui-c
 
 import WelcomeView from "./WelcomeView"
 
-const AppContent = () => {
+const AppContent = (): JSX.Element => {
   return (
     <>
       <Breadcrumb>
-        <BreadcrumbItem icon="home" label="Template App Home" />
+        <BreadcrumbItem onClick={() => {}} label="Template App Home" />
```
guoda-puidokaite, line 15: Following up from https://github.com/cloudoperators/juno/pull/673#discussion_r1881676818...
Can we keep this original? `icon="home"`
This would default to `help`, if not provided.

barsukov: I will rollback it too good catch Mrs :)

apps/template/src/App.tsx
guoda-puidokaite: Still there

.changeset/olive-ties-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudoperators/juno-ui-components": minor
+---
+
+Introduce types, depreacte other components, integration of new build system
```
guoda-puidokaite, line 5:
```suggestion
Introduce types, deprecate other components, integrate new build system
```

andypf (approved):

taymoor89 (approved):

Issue #529, opened 2024-10-17
[TASK](ui): rework Menu to become PopupMenu


franzheidl:
**Task Description**
The current Menu (and ContextMenu) component is flawed in several ways, and needs to be reworked as a more general `PopupMenu` component.

The new `PopupMenu` component should expose `PopupMenu.Toggle`, `PopupMenu.Menu`, `PopupMenu.Item` and `PopupMenu.Section` subcomponents.

The new `PopupMenu` component should accept a `toggle` and a `menu` as ~~props children. It should be possible to pass these as slotted props, or as~~ children:

```
<PopupMenu>
  <MyToggle />
  <MyMenu />
</PopupMenu>
```

Users can use subcomponents as exposed by `PopupMenu` (recommended), or their own custom components as `toggle` and `menu` respectively.

```
<PopupMenu>
  <PopupMenu.Toggle />
  <PopupMenu.Menu>
    <PopupMenu.Item />
  </PopupMenu.Menu>
</PopupMenu>
```

## Additional Requirements
* Both `toggle` and `menu` should be optional, resulting in the component rendering always.
* When no `toggle` is passed, `PopupMenu` should render an "overflow" / "kebap" icon as a toggle, allowing it to effectively replace the curren t`ContextMenu` component
* When no `menu` is passed, the toggle should still render and all necessary events should fire anyways, even when there is no menu to show.
* When passing children to the `PopupMenu`, the bevaviour should be as follows:
  * When subcomponents are being used, we render according to type / element, regardless of sequence (edge case: more than one `PopupMenu` and/or `PopupMenu.Menu` each: render the first one met, ignore the rest)
  * Clicking outside an open menu should close the menu

**Sub-tasks**

- [x] Create a new component in TS as a generic `PopupMenu`
- [x] Design a component architecture that wraps Headless Menu
- [x] Use sub components
- [x] Use the new PortalProvider to render the Menu into
- [x] Figure out and implement working, reliable positioning of the menu in the Portal
- [x] Make component render as an overflow menu per default, effectively replacing ContextMenu
- [x] Allow passing any Juno icon instead of default three vertical dots icon
- [x] Make sure all onClick handlers on custom component toggle still run
- [x] Allow passing the toggle as a child
- [x] Allow passing the menu as a child
- [x] Deprecate current Menu and ContextMenu components 
- [x] Make sure menu closes on click outside
- [x] Allow for rendering items as `div` or `anchor` 
[...]

PR #690, opened 2024-12-19
add PopupMenu

franzheidl:
# Summary

* Introduce a new `PopupMenu´ component to replace `Menu`(and related) as well as `ContextMenu` components. The new 

# Changes Made

- add `PopupMenu` component 
- deprecate `Menu`, `MenuSection`, `MenuItem`, `ContextMenu` components

# Related Issues

- #529 

# Screenshots (if applicable)

# Testing Instructions

1. `npm i`
2. `npm run test-vitest PopupMenu`
3. `npm run storybook`

# Checklist


packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+  // Test whether a PopupMenu.Menu child was passed:
+  const menu = childrenArray.find((child) => React.isValidElement(child) && child.type === PopupMenuMenu)
+
+  return (
+    <HeadlessMenu as="div" className={`juno-popupmenu ${className}`} {...props}>
+      {/* Update our open state when the open render prop from headless ui menu changes, so we can run the handlers when our tracking state changes: */}
+      {({ open, close }) => {
+        // Set our tracking open state outside of rendering cycle. Trying to set the state during render will cause an error:
+        useEffect(() => {
```
barsukov, line 235: This is a big part in our next refactoring sessions, we need to get rid of this unnecessary use effects, I believe that you could just avoid it here and use simple setState here.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  disabled = false,
+  icon = "moreVert",
+  menuSize = "normal",
+  onClose = undefined,
+  onOpen = undefined,
+  ...props
+}) => {
+  // Create a state to track headless-ui's internal open state (Note that we can track the internal state and can let headless ui update our tracking state, but we can not set it, hence there is no `open` prop on our component):
+  const [isOpen, setIsOpen] = useState(false)
+  const previousOpenRef = useRef(isOpen) // Track the previous value of isOpen for efficiency
```
barsukov, line 207: could you clarify why you need this ?

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  children = null,
+  className = "",
+  disabled = false,
+  icon = "moreVert",
+  menuSize = "normal",
+  onClose = undefined,
+  onOpen = undefined,
+  ...props
+}) => {
+  // Create a state to track headless-ui's internal open state (Note that we can track the internal state and can let headless ui update our tracking state, but we can not set it, hence there is no `open` prop on our component):
```
barsukov, line 205: so many comments here in PR in general, I do not see personally any values in them,  looks like generated content for me. Please, if it is possible remove them.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  if (!context) {
+    throw new Error(
+      "The usePopupMenuContext hook must be used inside within a PopupMenu that provides the PopupMenuContext."
+    )
+  }
+  return context
+}
+
+// Export the context for advanced use cases:
+export { PopupMenuContext }
```
barsukov, line 181: Could you please clarify why do you need a separated exported context?

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+// ----- Interfaces -----
+
+// Type the PopupMenu context
+export interface PopupMenuContextType {
+  close: () => void
+  isOpen: boolean
+  menuSize: "normal" | "small"
+}
+
+export interface PopupMenuProps extends React.ComponentProps<typeof HeadlessMenu> {
```
barsukov, line 107: (question)

Does the HeadlessMenu not providing any types? that's why you use typeof here?

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+}
+
+export interface PopupMenuToggleProps extends React.ComponentProps<typeof HeadlessMenu.Button> {
+  className?: string
+}
+
+//Extract types directly from React.ElementType as Headless Menu Items is a dynamically typed component that cannot be extended with React.ComponentPropsWithRef (such as `href` or `target` – these will be accepted when passing `as` as `a` in order to render an anchor, but not when rendering a div.):
+type HeadlessMenuItemsProps = React.ComponentPropsWithRef<React.ElementType>
+// Extend the extracted props instead:
+export interface PopupMenuMenuProps extends HeadlessMenuItemsProps {
```
barsukov, line 131: Could you please clarify naming **PopupMenuMenuProps** ? Why it is MenuMenu?

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  className?: string
+  disabled?: boolean
+  href?: string // accept a href when rendering a link, whether or not to pass it to HUI item will behandled in the component
+  icon?: keyof typeof KnownIconsEnum
+  label?: string
+  rel?: string // accept rel for anchors, handle in component
+  target?: string // accepot target for links, handle in component
+}
+
+export interface PopupMenuSectionProps {
```
barsukov, line 149: This component is quite big, I would suggest to extract all the types related code nearby in PopupMenu.types.ts file what do you think?

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+    >
+      {/* 
+      To render the item content, provide an itemBag context to the headless ui API. This includes properties like active, disabled, close. 
+      Then, if label is a string, render it. Otherwise, if children is a function, invoke it with itemBag; otherwise render children straightaway:
+    */}
+      {(itemBag) => (
+        <>
+          {icon && <Icon icon={icon} size="18" className={`juno-popupmenu-item-icon ${itemIconStyles}`} />}
+          {/* Make sure to test for truthy label and then check for string type, otherwise typeof label === string would alway return true due to the default empty string still being of type string: */}
+          {label && typeof label === "string"
```
barsukov, line 348: (Question)
I would not follow the guides in this case that we used to follow everywhere and try to think maybe in a advance here, since the component is new. 
Why should we support string ? why should we send children as a function? or other way around, is it the way to avoid this logic and checks? Open for discussion

packages/ui-components/src/components/MenuSection/MenuSection.component.tsx
```diff
@@ -18,8 +19,13 @@ const titleStyles = `
 	jn-text-xs
 	jn-p-2
 `
-/** Use MenuSection to structure and sub-divide MenuItems in a menu. All but the last MenuSection will render a visible divider at the bottom. Optionally, a MenuSection can have a title.*/
-export const MenuSection: React.FC<MenuSectionProps> = ({ title = "", children = null, className = "", ...props }) => {
+/** DEPRECATED: Use `PopupMenu` with `PopupMenu.Section` instead. Use MenuSection to structure and sub-divide MenuItems in a menu. All but the last MenuSection will render a visible divider at the bottom. Optionally, a MenuSection can have a title.*/
+const DeprecatedMenuSection: React.FC<MenuSectionProps> = ({
```
barsukov, line 23: (Question)
Would it possible to move this under a directory Deprecated?

apps/example/src/components/PositionArea.jsx
```diff
@@ -0,0 +1,80 @@
+import React, { useState } from "react"
+import PropTypes from "prop-types"
+import { Radio, RadioGroup, TextInput } from "@cloudoperators/juno-ui-components"
+
+/** A container component to test positioning and alignment behaviour of components */
+const PositionArea = ({ children = null, className = "", ...props }) => {
```
barsukov, line 6: Since we have already opened PR from @guoda-puidokaite on migration example APP to TS, maybe you could do all example app changes in a following PR? to avoid the conflicts?
WDYT?

apps/example/src/components/AppContent.jsx
```diff
@@ -89,6 +91,19 @@ const AppContent = () => {
                     <SelectOption value="4" label="Option 4" />
                   </Select>
                 </Container>
+                <PositionArea>
```
barsukov, line 94: For avoiding conflicts would be nice to have It in following PR.

barsukov (changes requested):
Great job, I have left som comments that could be Optional and open for discussion as (Question) marked
Sum up review and changes requested:
1) Please delete "generated" comments ;)
2) Some usage of useEffect, could be unnecessary
3) Extract types in separated file.
👍

packages/ui-components/src/components/ContextMenu/ContextMenu.component.tsx
```diff
@@ -103,3 +90,10 @@ export const ContextMenu: React.FC<ContextMenuProps> = ({
     </Menu>
   )
 }
+
+DeprecatedContextMenu.displayName = "ContextMenu"
+
+export const ContextMenu = withDeprecationWarning(
```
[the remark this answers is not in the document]
edda: I like this idea. But that's a bigger topic, not for this PR. Can you add it to the "improvements after migration" epic?

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
edda: I believe we mentioned this in the "improvements after migration" epic, right? Wouldn't mess with what we know works in the context of this PR

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
edda: I disagree. Especially for complex components like this that has to deal with/work around idiosyncrasies of the underlying headless library I'd rather err on the side of slightly too many comments. They help in months' or years' time when we have to try to understand why we do things a certain way.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
edda: Currently the standard is that types are defined in the same file as the component itself. I'm not against extracting them, but we should make a general decision and do it the same everywhere. We have always ensured to follow the same standard rules structurally with all components. We shouldn't break it in the context of this PR. Another one for #570

packages/ui-components/src/components/MenuSection/MenuSection.component.tsx
edda: Let's do this in a separate PR for all deprecated components at once.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
edda: I believe it's PopupMenu-Menu-Props. So the props for the menu component of `PopupMenu`. Makes sense, just looks a bit weird.

apps/example/src/components/PositionArea.jsx
guoda-puidokaite: Thanks, @barsukov! @franzheidl, feel free to merge your PR when ready, mine has already been merged.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: Headless Menu does not directly expose a state whether it is open or not, the only thing we can get is a render prop. In order to run our handlers and overall keep our APIs consistent as much as possible, I have introduced a state that tracks the internal state of Headless Menu, so we can read, add it to our context so we and users of our components can use it whenever and however they/we want. It is updated during the render cycle when Headless Menu's `open` render prop changes.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: I agree 100% with Esther here. This is a pretty complex component, and by wrapping Headless Menu we have to apply some approaches that are not obvious and far from being self explanatory. Not sure what makes you think these are generated though?

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: Users can import the context in order to read the open state of the menu, so they can apply their own styling on custom toggle elements depending on the open state of the menu, or implement any other styling or funcitonality depending on its state. The context also exposes a function to close the menu, so any arbitrary element in or outside the menu can be configured to close it. `PopupMenu`is designed as a rather generic component, so the menu could contain elements other than `PopupMenu.Item`.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: My understanding is this is the correct way to use the types provided by Headless Menu and extend them with our own? I am still new to TypeScript and this is by far the most complex component I have ever written in TS, I may be wrong here and happy to learn how if there is a better or more idiomatic way to do this.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: `PopupMenuMenu` is the subcomponent that is the actual menu element of `PopupMenu`. As `PopupMenu` aptly describes the component as a whole and what it does, I needed a name for the actual menu element. Headless internally calls this `MenuItems`, but as this looks almost identical to `MenuItem`(just missign the `s`) and I found this very confusing in the docs of Headless menu, I decided to call it something else. Also `MenuItems`suggests this is just a collection of items, but in reality it isn't, it is the part of the component that holds the most functionality/complexity, and it could easily be something else other than just a collection of `PopupMenu.Item` elements (see my above comment re exported context). `PopupMenuMenu` looks odd enough and is decidedly longer, I found it works better when scanning the code.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: To stay consistent with all our other components, we need to accept both string and children. A user may pass a function as children in order to make use of the render props (`itemBag`) Headless Menu provides per child (see comment in the code). Not handling this case will fail and produce errors.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: p.s. I have updated the "With Toggle As Button Styled By State" story to better illustrate the concept and how to use the context.

franzheidl:
Thanks so much for your time and comments, @barsukov, @guoda-puidokaite , @edda. 👍

I will agaon over all of these and further comment or resolve where possible.

I will also merge in `main` again and resolve any conflicts, and also update the `Default` story as requested by @edda to render a menu already, even thoguh the component does not have one by default. Just to avoid potential confusion.

Thanks again everybody! 🙏

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
franzheidl: Generally I do agree with using fewer `useEffect`s, we may have gone a bit overboard with these in the past, however this case is special:

In order to update our tracking state, we need to rely on the render props as provided by Headless menu. There is no other way of knowing whether the menu is open or not. This render prop updates in the components render cycle, using a useEffect as in an anonymous function that recieves the render props from headless menu in an object as an argument returned by the component, and again is used to update our tracking state. This is a valid (and the only afaik) way to achieve this, so this useEffect is crucial for the functionality of the component. (Also see next comment).

franzheidl:
Hey all, re-requesting review here. 

As to the requested changes – these are addressed in more detail in individual comments, here's the high-level:

Please delete "generated" comments ;)
**Leaving comment in, we believe them to hold enough value in the context of this component**

Some usage of useEffect, could be unnecessary
**We understand the general sentiment and agree 100%, not applicable in the context of this specific component/occuence

Extract types in separated file.
**General decision for all our components, not in scope of this component/PR (although it illustrates the point)**

barsukov (commented):
I leave it as a comment
https://github.com/cloudoperators/juno/pull/690/files#r1918881779
The only one thing is open for me and kinda blocker, but I still do not wanna block it cause of going of vacation, please reach out my colleges that I named for a better clarification and help on this topic 🥇

barsukov (commented):
Sorry, my bad just discovered that HeadlessMenu is not control by us :) so all good, maybe we could leave it as it is and you are right in this point delete all my previous comments.

barsukov (dismissed):

packages/ui-components/src/components/PopupMenu/PopupMenu.stories.tsx
```diff
@@ -0,0 +1,320 @@
+/*
```
guoda-puidokaite, line 1: **(Question)**

Is this suppose to happen for other component Stories? Can we disable or improve?

Screenshot 2025-01-17 at 14 04 10

guoda-puidokaite (commented):
Seems to work well in Storybook. 💪🚀 Thanks, @franzheidl.

I left some comments wondering if we could tidy and organise the code a little more, as the main component file was quite overwhelming to review.

I think there might be some undressed comments in the PR too. 

Let me know what you think!

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+      className={`juno-popupmenu-menu juno-popupmenu-menu-size-${menuSize} ${menuStyles} ${className}`}
+      {...props}
+    >
+      {children}
+    </HeadlessMenu.Items>
+  )
+}
+
+// ITEM COMPONENT
+const PopupMenuItem: React.FC<PopupMenuItemProps> = ({
```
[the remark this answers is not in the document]
barsukov: That also was my original suggestion over the types...

packages/ui-components/src/components/PopupMenu/PopupMenu.stories.tsx
edda: I noticed this, too. It looks like a new storybook feature to me where they extract subcomponents in the same file into their own tab. Definitely bears investigating but would also add this to #570 because it probably affects multiple components.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+`
+
+// Make sure the floating reference wrapper element around the toggle fits snug around the toggle, so the positioning of the menu is correct relative to the visible toggle:
+const floatingReferenceWrapperStyles = `
+  jn-inline-flex
+`
+
+// ----- Interfaces -----
+
+// Type the PopupMenu context
```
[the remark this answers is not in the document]
franzheidl: I have delete some of these rather obvious comments.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
```diff
@@ -0,0 +1,394 @@
...
+export interface PopupMenuSectionSeparatorProps {
+  className?: string
+}
+
+// ----- Define PopupMenu context and hook -----
+
+// Context:
+const PopupMenuContext = createContext<PopupMenuContextType | null>(null)
+
+// Hook to use the context:
```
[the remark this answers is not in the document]
franzheidl: See previous comment.

packages/ui-components/src/components/PopupMenu/PopupMenu.component.tsx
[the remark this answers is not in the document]
franzheidl: I have added the topic to #570 .

franzheidl:
Thank you, @guoda-puidokaite !

I have removed some of the comments as requested, but I really would like the others to remain intact in order to keep the component maintainable in the future. Some of the comments just help to structure the file, at least until we have made a decision as to to whether we want to extract subcomponents and / or types into individual files generally.

This decision is a general one and out of scope of this component, I have added the topic to #570 as suggested by @edda .

Also added the topic of making subcomponent props available via the new respective tabs in storybook to #570 , as if and how this can be done most likely depends on our decision as to how we want to split component or not. 

Thank you again, I would kindly ask you to re-review and, if you can, approve merging this.

edda (approved):

guoda-puidokaite (approved):

edda (approved):

guoda-puidokaite (dismissed):

guoda-puidokaite (approved):

edda (approved):

PR #700, opened 2024-12-20
fix(ui): ensure main and header container default for fullwidth behaviour is the same

edda:
# Summary

This PR ensures that the default fullWidth behaviour is the same for both the main container and the header container.

Default behaviour is meant to be as follows:
* In non-embedded mode, fullWidthContent should default to false, unless explicitly set to true.
* In embedded mode though, fullWidthContent should default to true, unless explicitly passed as false.

If using `AppShell` for layout the main container (which holds the app content) followed this logic. However the recently added `HeaderContainer` did not, it always defaulted to `false`. This PR ensures that they both behave in the same way when it comes to defaults for this prop.

# Changes Made

- Ensured fullWidth default is applied to both `HeaderContainer` and main container in `AppShell`
- Refactored `AppShell`: moved `HeaderContainer` to jsx section instead of rendering it via function
- Added tests to `AppShell` to ensure main and header container behave the same in all combinations of `embedded` and `fullWidthContent`
- Had to update `apps/template/App.tsx` because the typecheck was failing if `fullWidthContent` wasn't passed to `AppShell` <-- I don't understand this, the prop is marked as optional in the type ??!?

# Testing Instructions

1. `npm i`
2. `npm run test`
3. start an app with `TopNavigation`  and run it with `embedded = true` --> both header (i.e. top nav) and main app content should default to full width
4. run the app with `embedded = false` --> both header and main app content should default to max width rendering (i.e. not full width)
5. `embedded = true` + `fullWidthContent = false` --> same behaviour as 4.
6. `embedded = false` + `fullWidthContent = true` --> same behaviour as 3.

# Checklist

### DIFF
--- a/apps/template/src/App.tsx
+++ b/apps/template/src/App.tsx
@@ -55,6 +55,7 @@ export const App = (props: AppProps) => {
         embedded={props.embedded === "true" || props.embedded === true}
         sideNavigation={null}
         topNavigation={null}
+        fullWidthContent={false}
       >
         <ErrorBoundary fallbackRender={fallbackRender}>
           <AppContent />
--- a/apps/template/src/App.tsx
+++ b/apps/template/src/App.tsx
@@ -12,6 +12,7 @@ import { ErrorBoundary } from "react-error-boundary"
 interface AppProps {
   theme?: string
   embedded?: string | boolean
+  fullWidthContent?: string | boolean
   endpoint?: string
   currentHost?: string
 }
@@ -48,6 +49,16 @@ export const App = (props: AppProps) => {
     )
   }
 
+  // ensure that the fullWidthContent prop is a boolean or undefined if not passed
+  // TODO: this is a workaround to avoid passing a string to the AppShell component but it is ugly.
+  // We should find a better way to normalize the prop to boolean in some central location. The same applies to the embedded prop.
+  const calculatedFullWidthContentValue =
+    props.fullWidthContent === "true" || props.fullWidthContent === true
+      ? true
+      : props.fullWidthContent === "false" || props.fullWidthContent === false
+        ? false
+        : undefined
+
   return (
     <QueryClientProvider client={queryClient}>
       <AppShell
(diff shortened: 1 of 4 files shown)


edda:
Would be particularly interested in why the typecheck failed in `App.tsx` of the template app (see details above)

packages/ui-components/src/components/AppShell/AppShell.component.tsx
```diff
@@ -18,49 +18,32 @@ import { HeaderContainer } from "../HeaderContainer/index"
...
-      return null
-    }
-    return (
-      <HeaderContainer fullWidth={fullWidthContent === true}>
-        {pageHeader && typeof pageHeader === "string" ? <PageHeader heading={pageHeader} /> : pageHeader}
-        {topNavigation}
-        {/* Wrap everything except page header and footer and navigations in a main container. Add top margin to MainContainerInner as we are not in embedded mode here. */}
-      </HeaderContainer>
-    )
+  let fullWidthOrDefault
```
barsukov, line 32: What do you think about this syntax ? 
```
const fullWidthOrDefault = embedded 
  ? fullWidthContent !== false 
  : fullWidthContent === true;
```

barsukov (commented):
Good from my side, just one little thing :) maybe could work as an improvements.

apps/template/src/App.tsx
```diff
@@ -55,6 +55,7 @@ export const App = (props: AppProps) => {
         embedded={props.embedded === "true" || props.embedded === true}
         sideNavigation={null}
         topNavigation={null}
+        fullWidthContent={false}
```
ArtieReus, line 58: Was this setting intentional, or was it just a test? I was under the impression that our default app setup behavior should be:  
- **embedded = true** → content spans the full width.  
- **embedded = false** → content  in width-constrained, centred.  

By explicitly setting `fullWidthContent` to `false` in the template app, the content will always be constrained to a centered column with a limited width, regardless of the value of `embedded`.

ArtieReus (commented):
Tested PR with `supernova` testing all possible combinations successfully 🎉

I just have a comment on the hardcoded default value for `fullWidthContent ` in the template app.

apps/template/src/App.tsx
edda: Yes, true. Good catch. I didn't want to specify this prop at all in the template app but got a typecheck error when I left it out (see above). I guess what we should do instead is the same as for the `embedded` prop, i.e. use the value that was passed from outside if available, undefined otherwise.

ArtieReus (approved):
Lgtm

barsukov (approved):

Issue #705, opened 2025-01-03
[EPIC](greenhouse): Topics from the ORA Workshop (December 2024)


ArtieReus:
### Description

This epic addresses key topics and improvements for the Greenhouse project discussed during the ORA Workshop held in December 2024. The main focus areas include simplifying the token setup for development, refining image versioning in Helm charts, and enhancing the deployment pipeline for QA and production

### Objectives

- Token Setup for Development. Enable a token setup for development environments that does not require authentication. Evaluate making the setup applicable beyond just the demo organization by dynamically determining the organization from application properties or the URL.

- Image Version for Helm Charts. Use image versions instead of digests for Helm charts. Extend the pipeline to maintain support for image digest usage in continuous deployment to QA. Ensure UI deployments to production occur weekly, tagging the production image version with an additional tag that maps to the Greenhouse release version.

### Acceptance Criteria

**Token Setup for Development**
- Development environments can use tokens without authentication.
- The organization is dynamically read from application properties or the URL when using tokens.
- Token setup works for all organizations.

**Deployment**
- Helm charts reference image versions instead of digests.
- The pipeline still uses continuous deployment to QA using image digests.
- Production deployments occur weekly, with production image versions tagged with Greenhouse release versions.

### Tasks

**Token Setup for Development**
- [x] #708 
- [x] #714
- [ ] #706 
- [ ] Implement Token Setup without authentication and dynamically determining the organization to be used.

**Deployment**
- [ ] Adjust Helm charts to use image versions as default ensuring compatibility with image digest-based deployments for QA.

- [ ] Implement tagging for production images with Greenhouse release version mappings.
- [ ] Improve Trivy vulnerability scanner displaying table output and fetching db from keppel

PR #708, opened 2025-01-07
feat(oauth): create token session

ArtieReus:
# Summary

Implement token session management using token data, with options to extend or override attributes, returning the same interface as `oidcSession`. This approach avoids integrating a token within the `mockSession`, preventing confusion from mixing two distinct use cases.

# Changes Made

- Added a new function in typescript for `tokenSesion`
- Decodes token and parses the information
- Uses the same interface as `oidcSession`
- Adds tests

# Related Issues

- #705

# Testing Instructions

1. `npm i`
2. `npx turbo test --filter @cloudoperators/juno-oauth`

# Checklist


packages/oauth/src/tokenSession.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+
+interface TokenSessionReturnType {
+  login: () => void
+  logout: () => void
+  currentState: () => Record<string, any>
+}
+
+export default function tokenSession(params: any = {}): TokenSessionReturnType {
```
barsukov, line 40: I do not think that you need to specify params as any, please provide a type accodring to params that you expect. at least token, onUpdate, options and then you can use key syntax of other object related params

packages/oauth/src/tokenSession.ts
```diff
@@ -0,0 +1,73 @@
...
+    state = { auth: authData, error: null, loggedIn: true, isProcessing: false }
+    onUpdate(state)
+  }
+
+  const logout = () => {
+    state = { auth: null, error: null, loggedIn: false, isProcessing: false }
+    onUpdate(state)
+  }
+
+  login()
```
barsukov, line 66: you call login method and then return login as a function? could you clarify ?

packages/oauth/src/tokenSession.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  if (Object.keys(unknownProps).length > 0) {
+    console.warn(
+      `WARNING: (OAUTH) unknown options: ${Object.keys(unknownProps).join(",")}. Allowed options are token, onUpdate`
+    )
+  }
+
+  const authData = composeAuthData(token, options)
+  let state: Record<string, any> = { auth: null, error: null, loggedIn: false, isProcessing: false }
```
barsukov, line 54: do not get a bit with a state, so why create a variable then mutate it? maybe it should be const instead?

packages/oauth/__tests__/tokenSession.test.ts
```diff
@@ -0,0 +1,103 @@
+import { describe, it, expect, vi, beforeEach } from "vitest"
+import tokenSession, { composeAuthData } from "../src/tokenSession"
+
+// Mock the functions decodeIDToken and parseIdTokenData
+vi.mock("../src/tokenHelpers", async () => {
```
barsukov, line 5: we control the function, why do we need to mock it?

barsukov (changes requested):
I have left few comments for my clarification

packages/oauth/src/tokenSession.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export default function tokenSession(params: any = {}): TokenSessionReturnType {
+  const { token, onUpdate, options, ...unknownProps } = params || {}
+
+  if (typeof onUpdate !== "function") {
+    throw new Error("(OAUTH MOCK) onUpdate should be a function")
+  }
+
+  if (Object.keys(unknownProps).length > 0) {
```
andypf, line 47: You could also consider creating an options guard. The zod library is excellent for this purpose. It’s just a suggestion and not a blocking change. 🙂

```ts
import { z } from "zod";

// Define the schema
const optionsSchema = z.object({
  token: z.string().optional(),
  onUpdate: z.function().optional(),
});

// Validate unknownProps
function validateOptions(options: unknown) {
  const result = optionsSchema.safeParse(options);

  if (!result.success) {
    const unknownKeys = Object.keys(options).filter(
      (key) => !(key in optionsSchema.shape)
    );

    console.warn(
      `WARNING: (OAUTH) Unknown options: ${unknownKeys.join(", ")}. Allowed options are: ${Object.keys(optionsSchema.shape).join(", ")}`
    );
  }

  return result;
}
```

andypf (commented):
Great job! 
I appreciate that you followed the structure and added another session option. There are some suggestions from @barsukov, and I’ve also added a few comments that might help improve the code further. Keep up the great work!

packages/oauth/src/tokenSession.ts
ArtieReus: Sounds good. I didn't know about this lib, but makes sense

packages/oauth/src/tokenSession.ts
ArtieReus: Yes, you are right. This is a copy paste error...

packages/oauth/src/tokenSession.ts
ArtieReus: I just want to initialize the session by default using the login process. The login should also be returned in case the user logs out. I am following the interface of the other sessions besides login the user directly with the token.

packages/oauth/src/tokenSession.ts
ArtieReus: I will check it. As I mentioned already I took as a base the other sessions without thinking really on the state topic. Will check.

packages/oauth/__tests__/tokenSession.test.ts
ArtieReus: There is already tests for those functions. They are out of the scope of the session. That's why they are mocked.

packages/oauth/src/tokenSession.ts
ArtieReus: fixed

packages/oauth/src/tokenSession.ts
ArtieReus: fixed

packages/oauth/src/tokenSession.ts
ArtieReus: fixed with types

packages/oauth/__tests__/tokenSession.test.ts
ArtieReus: I revisit this question and I wouldn't like to now for example with default parameters should I expect from the parser function and introduce mock objects that match this defaults. I just want to concentrate on testing the `tokenSession` and `composeAuthData` without testing helpers which are already tested.

packages/oauth/src/tokenSession.ts
ArtieReus: fixed

andypf (approved):

packages/oauth/src/tokenSession.ts
andypf: Type guards are meant for runtime checks, not just development. During development, TypeScript already ensures type safety, but at runtime, type guards provide additional validation to ensure your program handles data correctly in dynamic scenarios. ;)

packages/oauth/__tests__/tokenSession.test.ts
barsukov: I am stil strongly against it, it is a sign of trying to solve a coupling problem here and avoiding it, creating a future problems... solution of it is a DI pattern, please check it more out

packages/oauth/src/tokenSession.ts
```diff
@@ -0,0 +1,94 @@
...
+  const login = () => {
+    onUpdate({ ...state, auth: authData, loggedIn: true })
+  }
+
+  const logout = () => {
+    onUpdate({ ...state, auth: null, loggedIn: false })
+  }
+
+  if (initialLogin) {
+    login()
```
barsukov, line 86: I think you should call this method by yourself separately and tokenSession should just return you a methods....

barsukov (changes requested):
I still have comments on this one... lets discuss them

packages/oauth/__tests__/tokenSession.test.ts
barsukov: I think you could just call this methods :)

andypf (approved):

packages/oauth/src/tokenSession.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+import { parseIdTokenData, decodeIDToken } from "./tokenHelpers"
+
+interface AuthData {
+  JWT: string
+  raw: Record<string, any>
+  refreshToken: string
+  parsed: any // Adjust this type as per the return type of parseIdTokenData
+}
```
guoda-puidokaite, line 13: **(Type Improvement)**

Could this work? Locally looks good for me. 

```suggestion
interface AuthData {
  JWT: string
  raw: Record<string, unknown>
  refreshToken: string
  parsed: ReturnType<typeof parseIdTokenData>
}
```

guoda-puidokaite:
Just reviewing

packages/oauth/src/tokenSession.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export const composeAuthData = (token: string, options: any): { authData: AuthData | null; error: Error | null } => {
+  let tokenData = {}
+  let parsedData = {}
```
guoda-puidokaite, line 17: **(Type Improvement)**

Could this work? Locally looks good for me. 

```suggestion
export const composeAuthData = (token: string, options: Record<string, unknown> = {}): { authData: AuthData | null; error: Error | null } => {
  let tokenData: Record<string, unknown> = {}
  let parsedData: unknown = {}
```

guoda-puidokaite (commented):
Type improvements. But other than that, looks great. 💪

packages/oauth/src/tokenSession.ts
ArtieReus: Make sense. I will do it in the next PR to not lose the approvals.

packages/oauth/src/tokenSession.ts
ArtieReus: Make sense. I will do it in the next PR to not lose the approvals.

Issue #711, opened 2025-01-10
[Task](url-state-provider): implement query string encoder and decoder


taymoor89:
**Task Description**

To make our URLs look more intuitive, predictable, and easier to craft, implement:

- An encoder that takes a JavaScript `object` and returns a query `string` that is safe to transmit over the internet.
  - The encoder should only accept a non-nested object that contains keys with values that are either primitive, `RegExp`, or an array of primitive or `RegExp`.
- A decoder that takes a query `string` and returns a JavaScript object with all escaped values unescaped and parsed back to the corresponding data type.

**Related Issues**
- [x] #683

PR #712, opened 2025-01-10
feat(url-state-provider): new way of encoding and decoding query string

taymoor89:
# Summary

This PR implements a new `encode` and `decode` function to persist a javascript object to the query string and parsing query string back to the same javascript object.

This whole idea of this solution has been verified by **Esther** and **Stefan Majewsky**(who needs to craft query string in this new way).

# Changes Made

- The new encode and decode methods that are exported as `encodeV2` and `decodeV2` are created inside `v2` directory to keep existing code intact.

# Related Issues

- [x] #711 

# Testing Instructions

1. `npm i`
2. `npm run test`

# Checklist


packages/url-state-provider/src/index.ts
```diff
@@ -324,3 +324,5 @@ export {
   decode,
   encode,
 }
+
+export { encode as encodeV2, decode as decodeV2 } from "./v2"
```
edda, line 328: Question:
I understand why you did this in the context of a minor update. But why not make it a new major version instead and have the `encode` and `decode` exports be the V2 versions?

taymoor89: The reason I did not created a major version just yet is because the new solution would run alongside the existing one so technically this is not a break change at the moment but after all apps are fully converted to the new URL structure then we can remove the old solution and release a major version as a breaking change and rename `encodeV2`, `decodeV2` exports. 
Moreover if we release a new major version we’ll have to make sure that any important change in the previous version(e.g vulnerabilities fix) will have to be separately released to the previous version as patch.

packages/url-state-provider/src/v2/encode.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const validateObjectToEncode = (object: ObjectToEncode) => {
+  if (object === null || typeof object !== "object" || object instanceof RegExp || Array.isArray(object)) {
+    return false
+  }
+
+  for (const key in object) {
+    const value = object[key]
+    if (!isPrimitive(value) && !Array.isArray(value)) {
+      return false
```
TilmanHaupt, line 28: So we are not able to nest Arrays in objects anymore and we need to refactor? Why not using superstate here?

packages/url-state-provider/src/v2/testCases.ts
```diff
@@ -0,0 +1,35 @@
...
+type TestCase = {
+  id: number
+  encoded: string
+  decoded: DecodedObject
+}
+
+const testCases: TestCase[] = [
+  {
+    encoded: "aString=someId1&aNumber=2&aBoolean=true&anArray=1,true,something",
+    decoded: { aString: "someId1", aNumber: 2, aBoolean: true, anArray: [1, true, "something"] },
```
TilmanHaupt, line 17: are stings like "null" "true" "[Infinity] encoded and decoded in the correct type? So far i do not understand how the decoding finds the correct type.

packages/url-state-provider/src/v2/encode.ts
```diff
@@ -0,0 +1,49 @@
...
+    typeof value === "number" ||
+    typeof value === "boolean" ||
+    value === null ||
+    value === undefined ||
+    value instanceof RegExp
+  )
+}
+
+const validateObjectToEncode = (object: ObjectToEncode) => {
+  if (object === null || typeof object !== "object" || object instanceof RegExp || Array.isArray(object)) {
```
ArtieReus, line 21: what about Your `empty objects` your validation currently allows empty objects (e.g., {})?

packages/url-state-provider/src/v2/encode.ts
taymoor89: Yes the rule is the following,
To make our lives simple, one can only supply flat js object to the `encode`, meaning no nesting, the only exception is that a value can also be an `Array` but that should only contain mixture of primitive types or `RegExp`. Check the types in `types.ts` file.
This is a completely different solution and ideally should not be mixed with superstate because we would not want to own any manual mapping rather rely on the standard js API `encodeURlComponent` to convert query strings to the URL safe characters.

packages/url-state-provider/src/v2/testCases.ts
taymoor89: I've updated that test case and added `null` as well. Yes please feel free to add/suggest more test cases.

The `query-string` package internally uses `decode-uri-component` package to type cast strings to the correct type. This is not the default behaviour but I enabled this by default in our wrapper package because I think this is a sensible default. Moreover we can also expose this settings from the `encode` to let programmer to turn this on/off on situation to situation basis.

packages/url-state-provider/src/v2/encode.ts
taymoor89: Good catch, I've fixed it and added a test case for that as well.

ArtieReus (approved):
Good Job! I like it. LGTM

packages/url-state-provider/src/v2/testCases.ts
taymoor89: While integrating this solution with supernova I've changed the default encoding behaviour of falsy values like `null` and empty strings `""` in a way that such fields will not be persisted in the URL which would save some space but this behaviour can easily be customised through encode `options`. See the [refactor commit](https://github.com/cloudoperators/juno/pull/712/commits/149c16f8f1638133fa8b6bc0140eb0b213ec1320).

TilmanHaupt (approved):

ArtieReus (approved):
re-approve

packages/url-state-provider/src/v2/decode.spec.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import decode from "./decode"
+import testCases from "./testCases"
+
+describe("decode", () => {
+  it.each(testCases)("[$id] should successfully decode given input", ({ encoded: input, decoded: output }) => {
```
barsukov, line 10: cam we do a proper testing and not make it implicit @taymoor89

packages/url-state-provider/src/v2/encode.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    expect(
+      encode({
+        key: "value",
+        null: null,
+        empty: "",
+      })
+    ).toBe("key=value")
+  })
+
+  it.each`
```
barsukov, line 24: I would really try to avoid it :) 
usually it will make failed test less readable and quite hard to see what is going on, test must be explicit

packages/url-state-provider/src/v2/encode.ts
```diff
@@ -0,0 +1,58 @@
...
+    typeof value === "boolean" ||
+    value === null ||
+    value === undefined ||
+    value instanceof RegExp
+  )
+}
+
+const validateObjectToEncode = (object: ObjectToEncode) => {
+  if (
+    typeof object === "undefined" ||
```
barsukov, line 22: we could discuss some tricks to avoid this casting parts using typescript tricks

packages/url-state-provider/src/v2/encode.spec.ts
taymoor89: Ah I've exactly the opposite opinion. I find `it.each` much more readable because this way you can find different scenarios/test cases all at once and can add a new one without repeating too much of the code. For example imagine converting the following to each individual test. Also if you have good description for each scenario it is quite easier to find which one is failing, so this is not a problem of `it.each` rather how one describes the test cases.

Screenshot 2025-01-14 at 14 02 56

PR #726, opened 2025-01-20
chore(ui): fix vite configuration for type output

guoda-puidokaite:
# Summary

**Issue Found**
- During the [migration of ui-components to Vite](https://github.com/cloudoperators/juno/pull/673/files), the types were configured to be outputted to `/build/types` instead of `/build`. 
- This caused "missing type" issues when components were used in certain apps. As a temporary workaround, some props were passed as `undefined` or `null`.

**Solution**
- This PR corrects the output path and removes the temporary fixes, instead of the proposed issue in the ticket.
- Additionally, since the correct types are now recognized in the apps, this PR also addresses the newly flagged type issues.

# Changes Made

- Corrected the type output path from `/build/types` to `/build`
- Removed props passed as `undefined` or `null` that were used as a temporary fix
- Fixed newly identified type issues in the apps

# Related Issues

- Main ticket https://github.com/cloudoperators/juno/issues/678
- Fixes issues introduced in https://github.com/cloudoperators/juno/issues/444

# Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have made corresponding changes to the documentation (if applicable).
- [x] My changes generate no new warnings or errors.

PR #735, opened 2025-01-22
chore(ui): add changeset for vite configuration fix to enable types in juno-ui-components

guoda-puidokaite:
# Summary

- Add changeset for https://github.com/cloudoperators/juno/pull/726/files
- Fix `Icon` display name

# Checklist

### DIFF
--- a/packages/ui-components/src/components/Icon/Icon.component.tsx
+++ b/packages/ui-components/src/components/Icon/Icon.component.tsx
@@ -892,7 +892,7 @@ export const Icon = forwardRef<HTMLAnchorElement | HTMLButtonElement, IconProps>
   }
 )
 
-Icon.displayName = "IconTs"
+Icon.displayName = "Icon"
 
 export interface IconProps
   extends Omit<React.HTMLProps<HTMLAnchorElement> | React.HTMLProps<HTMLButtonElement>, "size"> {
--- a/packages/ui-components/src/components/Icon/Icon.component.tsx
+++ b/packages/ui-components/src/components/Icon/Icon.component.tsx
@@ -837,62 +837,70 @@ const getColoredSizedIcon = ({ icon, color, size, title, iconClassName, ...iconP
   }
 }
 
-export const Icon = forwardRef<HTMLAnchorElement | HTMLButtonElement, IconProps>(
-  (
-    { icon = null, color = "", size = 24, title = "", className = "", href = "", disabled = false, onClick, ...props },
-    ref
-  ) => {
-    // if href or onClick was passed, then we want to add the passed classes and passed arbitrary props to the button or anchor
-    // otherwise add the passed classes/props to the icon itself
-    const iconClassName = href || onClick ? "" : className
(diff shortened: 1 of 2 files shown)


andypf (approved):
👍

packages/ui-components/src/components/Icon/Icon.component.tsx
```diff
@@ -892,7 +892,7 @@ export const Icon = forwardRef<HTMLAnchorElement | HTMLButtonElement, IconProps>
   }
 )
 
-Icon.displayName = "IconTs"
+Icon.displayName = "Icon"
```
ArtieReus, line 895: I was wondering if it’s possible to set the display name dynamically instead of hardcoding it. For example:
```js
Icon.displayName = Icon.name
```
or even with a fallback
```js
Icon.displayName = Icon.name || "Icon"
```

andypf:
Preview works!

packages/ui-components/src/components/Icon/Icon.component.tsx
andypf: I think this goes beyond the scope of this PR. Let’s focus on the changeset for now and discuss topics like dynamic or static approaches later. Here’s something to read in the meantime: [MDN: Function.displayName](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/displayName).

ArtieReus: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/displayName#changing_displayname_dynamically

ArtieReus: The comment is meant to verify if it makes sense and, if so, to create a task for the future.

andypf (dismissed):
Great, you changed it to function 💯

andypf:
@guoda-puidokaite there are issues with lint and ts checks :(

ArtieReus (dismissed):
I see a warining here as @andypf commented:
https://github.com/cloudoperators/juno/pull/735/files#diff-ebe4a6dc10bd9086daab8e5bd1252c3c9cca409668572c7477e14e30a4f425b3
`Unsafe assignment of an error typed value`

guoda-puidokaite:
Thank You for the review, @andypf @ArtieReus. I'm in the middle of making the suggested changes. 

No need to block the PR for pipeline errors, as they're already blocking the merge. 🙏

I'll convert to draft until done. 🚀

ArtieReus:
> Thank You for the review, @andypf @ArtieReus. I'm in the middle of making the suggested changes.
> 
> No need to block the PR for pipeline errors, as they're already blocking the merge. 🙏
> 
> I'll convert to draft until done. 🚀

You are right!! I didn't left the code tab thats why I didn't see it 😅

andypf (approved):

hodanoori (approved):

edda (approved):

Issue #664, opened 2024-12-05
[Task](context): Double check types


guoda-puidokaite:
**Task Description**
TBC

**Sub-tasks**

- [ ] Sub-task 1
- [ ] Sub-task 2
- [ ] Sub-task 3

**Related Issues**
https://github.com/cloudoperators/juno/issues/168

**Additional Context**
Add any other context or information here.

guoda-puidokaite:
Please read PR descriptions of actual changes made:

- https://github.com/cloudoperators/juno/pull/748
- https://github.com/cloudoperators/juno/pull/759

PR #748, opened 2025-01-29
chore(ui): improve component and story types, move deprecated components (1st iteration)

guoda-puidokaite:
# Summary

Improved ui-component types, their Storybooks and moved deprecated components (1st iteration).

# Changes Made

- Moved deprecated `ContextMenu`, `Menu`, `MenuItem`, `MenuSection` to `deprecated_js` and didn't check the types of these components.
- Changed `activeItem` in `Navigation`, `SideNavigation`, `TabNavigation` and `TopNavigation` to be of type `ItemKeyType` (`ReactNode`)
- Checked and improved component types:
  - set `children` to be `ReactNode`, unless specific restrictions were required.
  - removed type `null`, it's not required unless there's a specific need to distinguish between a prop being `undefined`/not set vs. explicitly `null`.
  - changed `IntroBox` default values for `title`, `text` and `heroImage` from `null` to `""`.
  - removed unnecessary `null` and `undefined` default values.
  - **removed `object` from the `children` prop in `CodeBlock`, as it's [not valid](https://github.com/cloudoperators/juno/pull/748#discussion_r1940764241) to pass a plain JavaScript object as `children` in React. We should stick to valid React nodes.** 
- Checked that Storybook types match component types
  - set `children` to `ReactNode` where the type was missing.
  - added **missing** props to component Storybooks e.g. `children` was missing in `Button`.
  - removed **extra** props from component Storybooks that didn't exist in the components e.g. `name` doesn't exist in `ButtonRow`.
- Added a changeset.

# Related Issues

- https://github.com/cloudoperators/juno/issues/664

# Checklist


packages/ui-components/src/components/Modal/Modal.component.tsx
```diff
@@ -241,23 +241,23 @@ export interface ModalProps extends Omit<React.HTMLProps<HTMLDivElement>, "size"
...
   /** Pass an icon name to show on the cancelling button */
-  cancelButtonIcon?: KnownIcons | null
+  cancelButtonIcon?: KnownIcons
   /** Whether the modal will be open */
   open?: boolean
   /** The children of the modal. These will be rendered as the modal content. To render custom buttons at the bottom, see `modalFooter` below.*/
   children?: React.ReactNode
   /** Optional. Pass a `<ModalFooter />` component with custom content as required. Will default to using the `<ModalFooter/>` component internally. */
-  modalFooter?: JSX.Element | null
+  modalFooter?: JSX.Element
```
taymoor89, line 252: shouldn't we also change this to `ReactNode`? 🤔

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.component.tsx
```diff
@@ -23,7 +23,7 @@ export interface BreadcrumbItemProps extends React.HTMLAttributes<HTMLSpanElemen
   /**
    * The icon type to display in the breadcrumb item.
    */
-  icon?: KnownIcons | null
+  icon?: KnownIcons
```
taymoor89, line 26: I like the removal of `null`s in multiple places but I'm thinking if this could be a breaking change in case a consumer is relying on it and supplying `null` for some reason.

taymoor89 (commented):
Would it be possible to export `ItemKeyType` and `KnowIcons` types from `index.ts`? It is fine if you're planning to do it in the next iteration.

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.component.tsx
guoda-puidokaite: Glad to hear. 🙈 Only random props allow `null` explicitly, but they don't use it in any meaningful way in the code.

We want to make such type improvements (breaking type changes) now, before apps start using and really relying on them. For JS apps, it should still be okay?

packages/ui-components/src/components/Modal/Modal.component.tsx
guoda-puidokaite: Thank You so much for the thorough review! 🙏 

Most likely. It's on my list of things to address with @edda and then change in the next PR. For such cases, I want to first clarify if we only want to allow one JSX/React Element or multiple of a broader type (`ReactNode`).

guoda-puidokaite:
> Would it be possible to export `ItemKeyType` and `KnowIcons` types from `index.ts`? It is fine if you're planning to do it in the next iteration.

I'm guessing these types would help you for now so I did it. 😊 But yes, will address the rest once we're really confident about them.

Also about `ItemKeyType`: It's actually just an alias for `ReactNode` and not really a custom type. A little redundant? Though there's something nice about reusing this type in many components for the same prop, it gives a little more context. Wdyt?

packages/ui-components/src/components/NavigationItem/NavigationItem.component.tsx
```diff
@@ -4,11 +4,11 @@
...
 
 import React, { useContext, useEffect, useState } from "react"
+
 import { NavigationContext } from "../Navigation/Navigation.component"
-import { Icon } from "../Icon/index"
-import { KnownIcons } from "../Icon/Icon.component"
+import { Icon, KnownIcons } from "../Icon/Icon.component"
 
-type ItemKeyType = string | React.ReactNode | null
+export type ItemKeyType = React.ReactNode
```
guoda-puidokaite, line 11: Redundant, hmmm?

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.component.tsx
taymoor89: Yeah there are just very few internal `ts` consumers for this package so yeah it is fine.

packages/ui-components/src/components/NavigationItem/NavigationItem.component.tsx
taymoor89: I also think so, we can get rid of this in next iteration.

taymoor89 (dismissed):
LGTM 🚀

packages/ui-components/src/components/NavigationItem/NavigationItem.component.tsx
guoda-puidokaite: Done in this one since I already touched that type anyway 🙈 and don't want to export it.

guoda-puidokaite: Thank Youuu!

taymoor89 (dismissed):

taymoor89 (approved):

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
```diff
@@ -184,7 +184,7 @@ export interface CodeBlockProps extends Omit<React.HTMLAttributes<HTMLDivElement
   /** The content to render. Will override children if passed. */
   content?: string | object
   /** The children to render. Will be overridden by content prop if passed as well.  */
-  children?: React.ReactNode | object
```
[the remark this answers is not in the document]
taymoor89: Waiting for Esther to confirm this change.

taymoor89 (changes requested):

packages/ui-components/src/components/CodeBlock/CodeBlock.component.tsx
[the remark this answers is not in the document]
edda: Agreed, doesn't make sense to allow `object` as children and I think would also result in a RuntimeError. Good catch!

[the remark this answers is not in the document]
guoda-puidokaite: In that case I also need to review the component logic to make sure we're not passing context physically as children too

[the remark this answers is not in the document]
guoda-puidokaite: Niceee, doing it. 💪 Thank You!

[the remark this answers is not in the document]
guoda-puidokaite: Done. Only fixed the `children` logic in the component, not touching anything else.

taymoor89 (approved):

barsukov (approved):
LGTM

taymoor89 (approved):

Issue #761, opened 2025-02-10
[Task](Heureka): Integrate graphql backend


taymoor89:
**Task Description**
Integrate GraphQL backend.

**Sub-tasks**

- [ ] Generate types from GraphQL schema.
- [ ] Use Apollo Client to fetch data from the backend.
- [ ] List `Services` as a working demo.

**Related Issues**
Link to related issues or tasks, if any.

**Additional Context**
Add any other context or information here.

PR #779, opened 2025-02-14
feat(heureka): integrate GraphQL backend

taymoor89:
# Summary

This PR configures `graphql-codegen` to generate types and hooks from GraphQL Schema as well as configures `ApolloClient` to communicate with the GraphQL backend.

# Changes Made

- `graphql-codegen` configured.
- `ApolloClient` configured.
- Lists services in raw form to demonstrate successful communication.

# Related Issues

- #761 

# Screenshots (if applicable)

Screenshot 2025-02-14 at 15 17 41

# Testing Instructions

1. `npm i`
2. `npm run dev`

# Checklist


hodanoori (commented):
Thanks Taimoor. I'm wondering why preview of heureka-next is not loading for me whereas the heureka preview is coming up. Is it the same for you?

taymoor89:
> Thanks Taimoor. I'm wondering why preview of heureka-next is not loading for me whereas the heureka preview is coming up. Is it the same for you?

Yes I was also facing the same issue, the problem was that there was no `apiEndpoint` defined. The script was unable to generate `appProps.json` file because we do not have `HEUREKA_NEXT_APP_PROPS_BASE64` secret configured. For the moment I've reused `HEUREKA_APP_PROPS_BASE64` secret because it is essentially the same but I'll talk to Arturo to help us create one.

hodanoori (dismissed):
lgtm

apps/heureka-next/.env.template
```diff
@@ -0,0 +1,6 @@
+#
+# SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+
+API_ENDPOINT=SOME_URL
```
barsukov, line 6: end of line

apps/heureka-next/src/components/Shell/Shell.test.tsx
```diff
@@ -8,6 +8,14 @@ import { render, screen } from "@testing-library/react"
 import userEvent from "@testing-library/user-event"
 import { Shell } from "./Shell"
 
+/**
+ * let's mock Services because that is a dependency of Shell
+ * and has been tested independently
+ **/
+vitest.mock("../Services/Services", () => ({
```
barsukov, line 15: Why do we need to mock it? maybe we just do not need a test like that...

barsukov (changes requested):
good job for all of us! ;) 
I think we should not mock Services on tests of The shell, just write a simple tests for services...
other than that looks perfect

apps/heureka-next/src/components/Shell/Shell.test.tsx
taymoor89: As specified in the code comment, the reason to mock `Services` here is because it is a separate component and tested individually. Moreover in terms of unit tests, `Shell`'s responsibility is to only allow switching among different views but not how they look like.
If we do not mock `Services` then to render it properly we'll have to mock the graphql API requests that it makes(which has already been done in its own unit tests) that also means when adding more views to the `Shell` its unit tests will run a lot slower because it has to render everything that does not even belong to its implementation.

barsukov: I disagree with this statement I would rather prefer to remove the Shell tests itself then mock you own code. Cause this could be a very missleading in a way.

> 
If we do not mock Services then to render it properly we'll have to mock the graphql API requests that it makes(which has already been done in its own unit tests) that also means when adding more views to the Shell its unit tests will run a lot slower because it has to render everything that does not even belong to its implementation.

which will be the right way to do it...
I think the right way to do it to have a router component that could be act as a entry point and test only simple fallback entry instead.

barsukov: Also what I just realize, you are breaking the contract here, 
The Shell.tsx file has no knoledge about the Services and no connection at all here, so why you then mock explicity there in a tests? 
So please change it.

taymoor89: I disagree with removing `Shell` tests altogether because there is no router at the moment and that would also mean that we are leaving some logic untested. I also disagree that we should not mock our own code, we certainly can where it makes sense and makes tests run faster and that is exactly what I explained that the `Services` functionality is already tested in its own test file. For shell `Services` is a dependency and isn't its concern/implementation therefore it is fine to mock it. 

Also when writing tests using `react-testing-library` it is recommended to write tests thinking how a real user would use the UI and if you think about the user's interaction all the shell is doing is allowing user to switch among different views but not how they look that's the job of a view like `Services` and should be tested in its own sphere.

Note that the whole discussion is happening in terms of unit tests, for integration tests I'd suggest that we have e2e tests for important features as the integration tests are usually slower because they need to resolve a lot of dependencies which increase the time it takes to render a component and mostly result in timeout error or a random flakiness. We learned it in our previous project where we had to purge most of the integration tests and rather relying on unit and e2e tests to make our tests run faster on CI and even locally.

barsukov: Do you not see that you wrote this thing against SOLID principle and broke a contract here?

taymoor89: Ah instead of `Shell` it should be `ShellContent` but all of the discussion stays the same, as in we still need to mock `Services` in `ShellContent` tests because otherwise we’ll have to provide mock graphql response for `Services` or any other view that is not even a concern of `ShellContent` all it does is allowing switching among different views. 
And my comments are quite inline with even the first `Single responsibility` principle of SOLID where each component is doing it is own job and same should happen in the tests as well where a component should not worry about rendering its dependencies rather should focus on testing its own behaviors/implementation. And IMO this holds very true specially when a dependent component is making api requests and is already individually tested.

barsukov: > principle of SOLID where each component is doing it is own job and same should happen in the tests as well where a component should not worry about rendering its dependencies rather should focus on testing its own behaviors/implementation. And IMO this holds very true specially when a dependent component is making api requests and is already individually tested.

That's why the solutions like DI(Dependency Injection, Dependency Inversion or Inversion of Control) exists to solve this particular problem in case, that's what I mentioned it. If you start doing this things in tests mock around staff because "it should be tested somewhere else" you will land in a very funky spot later, trust me. 

And tests should be treated in same way as a code, it is reflection on your code and if you see this problem start appearing over and over agian, you should ask yourself if you are doing a right thing.

In this case as I said I would really make a change in a code to have ability to pass the property mock to shell content and shell itself and do it instead of mocking it.

So one og possible solution to create a option in a component called it mock and pass it as a flag, when it is enabled you do not render the ShellContent and boom you could avoid vi.mock.

edda: Right, so where to begin. In general I am also always wary of mocks because they can hide issues that result from later code changes that aren't included in the mock and then only become apparent at runtime. I tend to prefer to use `spy` when possible. 

In this particular case I actually find the use of mock quite elegant as we really don't care at all about what the Services view renders, only that it is rendered by default. So by mocking the view completely we remove the dependency to the Services view for the purpose of the test and thereby also ensure that this unit test won't break if changes are made to the Services view later that remove the thing we were testing for in this test. It's a neat separation of concerns.

However my question is: Should the decision which view to render by default by made in the `Shell` component or wouldn't it make more sense to do this in `ShellContent` where the logic that switches the views lives. There's even a switch statement there where the `default` case returns `null`. It could return `<Services />` instead. Then the whole render logic could be neatly unit tested there, probably with a spy instead of a mock as described below and the "renders Services by default" test could be removed from the `Shell` tests.

If you want to keep it as now I think it would still be possible to implement the test with a spy instead of a mock if the switch statement that decides which view to render in the `ShellContent` component was a separate function or maybe a hook. Then you could spy on it being called with `selectedView: SERVICES` and it would enable unit tests for the decision logic in ShellContent in general. This would probably have been my solution but maybe let's discuss face to face instead of wall-of-text style? ;)

taymoor89: Thanks for your opinion Esther yeah spy makes sense and I completely agree to have a conversation on such things rather than long text which sometimes is harder to interpret correctly. 
So we talked together with Wowa and Andreas and we think that we may not need to have separate `Shell` and `ShellContent` which was done for the readability purposes. Also because other people may start mocking things without thinking so we’ve agreed to wrap `Shell` around `MockProvider` contrary to what I was suggesting which would allows us to spy on api requests and apply data mocks just like in `Services.test.tsx`.

barsukov (approved):
LGTM!

hodanoori (approved):
lgtm

PR #780, opened 2025-02-16
refactor(supernova): fix all typescript migration leftovers for Regions and Alerts

### LINKED ISSUE #765: (Supernova): Fix all typescript migration leftovers for Regions
**Fix all typescript migration leftovers for Regions**

**Sub-tasks**
Please have a look on all regions/* folder and fix related files, this is aprx, files and count of errors:

- [x]  3  src/components/regions/Region.tsx:37
- [x]  3  src/components/regions/RegionsList.tsx:25

**Additional Context**
Sometimes you will need to fix few other staff, try to minimaze it,  keep PRs small and easy to review.

TilmanHaupt:
# Summary

Typed filter store and components, removing @ts-ignore and any from this part of the codebase. This is part of the post-Supernova migration effort to clean up TypeScript errors. More PRs will follow for other areas.

# Changes Made

closes #765 

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist


apps/supernova/src/lib/createAlertsSlice.tsx
```diff
@@ -6,19 +6,65 @@
...
   severityCountsPerRegion: {}, // {"eu-de-1": { total: number, critical: {total: number, suppressed: number}, warning: {...}, ...}
   regions: [], // save all available regions from initial list here
   regionsFiltered: [], // regions list filtered by active predefined filters
   enrichedLabels: ["status"], // labels that are enriched by the alert worker
   updatedAt: null,
 }
 
-// @ts-expect-error TS(7006) FIXME: Parameter 'set' implicitly has an 'any' type.
-const createAlertsSlice = (set, get) => ({
+const createAlertsSlice = (set: any, get: any): AlertsSlice => ({
```
barsukov, line 67: do this types need to be any? maybe we can find out how to solve it?

apps/supernova/src/lib/createAlertsSlice.tsx
```diff
@@ -96,7 +142,7 @@ const createAlertsSlice = (set, get) => ({
               // if it doesn't match, set visible to false and break out of the loop
               activePredefinedFilter &&
                 Object.entries(activePredefinedFilter.matchers).forEach(([key, value]) => {
-                  // @ts-ignore
+                  if (typeof value !== "string") return
```
barsukov, line 145: Do we need this guard? if we give the TS already knoledge about matchers maybe we will not need this casts

barsukov (commented):
few little staff but in a file itsleft of createAlertSlices we still have a lot of anys do you want to get reed of them later?

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: Thats because:   activePredefinedFilter: string | null - I don't know if "" is possible instead of null let me check.

TilmanHaupt: Holy crap. Thank you so much for seeing this. There is a huge logic problem which just worked because we only had one active PredefinedFilter till now.

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: I can do some research.

apps/supernova/src/lib/createAlertsSlice.tsx
```diff
@@ -5,20 +5,67 @@
...
+  setRegionsFiltered: (regions: string[]) => void
+  updateFilteredCounts: () => void
+  getAlertByFingerprint: (fingerprint: string) => AlertItem | undefined
+}
 
 const initialAlertsState = {
   items: [],
   itemsFiltered: [],
-  totalCounts: {}, // { total: number, critical: number, ...},
+  totalCounts: { total: 0, critical: 0 }, // { total: number, critical: number, ...},
```
edda, line 60: Why do you do a partial initialization here? The way it reads now it suggests that only the `total` and `critical` keys exist, but that's not true. The other severities are also counted (`warning` and `info`). So if you add initialization here it should be for all otherwise it might be confusing

apps/supernova/src/lib/createSilencesSlice.tsx
```diff
@@ -89,142 +124,139 @@ const validateTemplates = (templates: any) => {
...
-      /* 
+        /* 
       Given an alert fingerprint, this function returns all silences referenced by silencingBy. 
       */
-      getMappingSilences: (alert: any) => {
-        if (!alert) return
-        const externalSilences = get().silences.items
-        let silencedBy = alert?.status?.silencedBy || []
+        getMappingSilences: (alert) => {
+          if (!alert) return []
```
edda, line 172: Have you verified that changing the return to empty array doesn't break anything? If the place where the mapping silences are consumed doesn't explicitly check for empty array this might cause strange behaviour

apps/supernova/src/lib/createSilencesSlice.tsx
```diff
@@ -89,142 +124,139 @@ const validateTemplates = (templates: any) => {
...
+        },
 
-      /*
+        /*
       Find all silences (also expired ones) in items that matches all labels (key&value) from the alert but omit the labels that are excluded (excludedLabels)
       */
-      getSilencesForAlert: (alert: any) => {
-        if (!alert) return
+        getSilencesForAlert: (alert) => {
+          if (!alert) return []
```
edda, line 215: Similar as above, does changing the return value to empty array not break anything?

edda (changes requested):

apps/supernova/src/lib/createSilencesSlice.tsx
TilmanHaupt: All occurrences handle empty []

apps/supernova/src/lib/createAlertsSlice.tsx
```diff
@@ -5,6 +5,56 @@
...
+  regionsFiltered: string[]
+  enrichedLabels: string[]
+  updatedAt: number | null
+  actions: AlertsActions
+}
+
+export interface AlertItem {
+  fingerprint: string
+  labels: Record<string, string>
+  [key: string]: any
```
edda, line 31: I find this anonymous array pretty confusing. It's hard to know what this is used for. Can we make this in any way clearer?

apps/supernova/src/lib/createAlertsSlice.tsx
```diff
@@ -5,6 +5,56 @@
...
+}
+
+export interface AlertCounts {
+  total?: number
+  critical?: number
+  [key: string]: number | undefined
```
edda, line 37: Similar here with the anonymous key and it comes back to my earlier comment regarding initial value of `AlertCounts`. If we do expose some of the possible keys in the interface then we should do it for all the ones we know about. So we should add `warning?: number` and `info?: number` at least. I'm guessing the anonymous key is there for keys that we don't know about beforehand, right?

edda (changes requested):

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: I don't know there are a lot of fields from the API / k8s which may change over time

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: Your right. I just put the stuff from the tests in.

apps/supernova/src/lib/createAlertsSlice.tsx
```diff
@@ -5,6 +5,56 @@
...
+}
+
+export interface SeverityCounts {
+  total: number
+  critical?: { total: number; suppressed: number }
+  warning?: { total: number; suppressed: number }
+  [key: string]: any
```
edda, line 44: As above, if we do `critical` and `warning` we should also do `info`.

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: maybe just:  Record<[key: string]: number> ?

apps/supernova/src/lib/utils.ts
```diff
@@ -57,7 +57,7 @@ export const humanizeString = (value: any) => {
 //   }, ...
 // }
 export const countAlerts = (alerts: any) => {
-  const counts = { global: { total: 0 }, regions: {} }
+  const counts = { global: { total: 0, critical: 0 }, regions: {} }
```
edda, line 60: Sorry, I overlooked this one earlier. If we initialize the counts with critical we should also initialize `warning` and `info`

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: Such things will be a lot inside because esp. in alerts i often don't know the exact stuff. Maybe we should not add that then? Or i take a look in the k8s stuff

apps/supernova/src/lib/utils.ts
TilmanHaupt: Thats just used by the tests.

apps/supernova/src/lib/createAlertsSlice.tsx
edda: This isn't k8s, it's Alertmanager API. But yeah, it's fine. I actually preferred the earlier implementation where you specifically had `fingerprint` and `labels` in the interface because those are the important props on the item that we use all the time.

edda (approved):

apps/supernova/src/lib/createAlertsSlice.tsx
```diff
@@ -5,6 +5,59 @@
 
 import { produce } from "immer"
 import { countAlerts } from "./utils"
+import { Filter } from "./createFiltersSlice"
+import { AppState } from "../components/StoreProvider"
+import { StateCreator } from "zustand"
```
barsukov, line 10: Libs imports usually higher than the component/app imports.
Optional to change

apps/supernova/src/lib/createGlobalsSlice.tsx
```diff
@@ -3,57 +3,77 @@
...
+          set(
+            (state: any) => ({
+              globals: { ...state.globals, isURLRead: true },
+            }),
+            false
+          ),
+
+        setShowDetailsFor: (alertID) =>
+          set(
+            (state: any) => ({
```
barsukov, line 57: Should all any's still be there?

barsukov (commented):
great job little comments left

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: What do you mean?

apps/supernova/src/lib/createGlobalsSlice.tsx
TilmanHaupt: Deleted them now. Forgot because this slice was not in ticket scope - so it was just changed bc of options import.

TilmanHaupt: pushed.

apps/supernova/src/lib/createAlertsSlice.tsx
TilmanHaupt: ahh i understand

apps/supernova/src/lib/createAlertsSlice.tsx
ArtieReus: I would suggest to double check the `alertItem` types to review the alert raw data, which we display in the ui.

ArtieReus: and also the api:
https://github.com/prometheus/alertmanager/blob/main/api/v2/models/alert.go#L42

apps/supernova/src/lib/createGlobalsSlice.tsx
```diff
@@ -3,57 +3,77 @@
...
-          (state: any) => ({
-            globals: { ...state.globals, activeSelectedTab: activeSelectedTab },
-          }),
-          false,
-          "globals/setShowDetailsFor"
-        ),
+const createGlobalsSlice: (options?: Record<string, any>) => StateCreator<AppState, [], [], GlobalSlice> =
+  (options) => (set, get, store) => ({
+    globals: {
+      embedded: validateEmbedded(options?.embedded),
```
taymoor89, line 39: Thought it is slightly unrelated but because `validateEmbedded` is accepting `embedded` as `any` I'd suggest that we remove this method altogether and achieve the _similar_ result by typecasting `embedded` to `boolean` inline using any of the following ways.

```
embedded: !!options?.embedded
```
or 
```
embedded: Boolean(options?.embedded)
```

TilmanHaupt: Thats a good idea - but I, and the epic, wanted to keep all parsers and validators till i reshaped every slice to keep small pr. And then remove them.

edda (approved):

taymoor89 (approved):

PR #809, opened 2025-03-11
feat(greenhouse): pluginAdmin filtering

### LINKED ISSUE #754: [Feature](greenhouse-org-admin): Filtering for Plugin List
**Description**
We would like to add the possibility to filter the Plugin list within the greenhouse org-admin Area

**Problem Statement**
Currently we have an unsorted, unfilterable list of Plugins within the org-admin area. With increasing number of deployed plugins it becomes more and more unusable. E.g. I like to highlight `clusterNames` via the browser search to be able to focus on a set of Plugins

**Proposed Solution**
Add sorting and filtering options to the list.
Filter by:
- `Plugin.Spec.ClusterName`
- `Plugin.Spec.PluginDefinition`
- `Plugin.Status.StatusConditions.Ready = true/false/unknown`
- `Plugin.Metadata.Label[key] = [value]`

Sort by:
- `Plugin.Spec.DisplayName` (default)
- `Plugin.Spec.ClusterName` then `Plugin.Spec.DisplayName`

**Additional Context**
The `Plugin` list screen is going to be replaced by a list/view only showing `PluginPresets`. So this should just be a quick solution. Maybe *sorting* is superflous if default sorting by `displayName` is implemented.

TilmanHaupt:
# Summary

Closes #754 

1. The list is sorted in ascending order.
2. Search and regex apply to the entire pluginConfigItem element.
3. Filter by status.
4. Filter by label values, which are parsed.

# Testing Instructions

Filter & search in ORG > PLUGINS

# Checklist

# PR Manifesto


apps/greenhouse/src/components/core-apps/org-admin/components/plugins/plugins/components/Toolbar.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const handleValueChange = (value?: string | number | string[] | undefined) => {
+    setSelectedValue(value !== undefined ? String(value) : "")
+    setLabelValueFilter({ label: selectedLabel, value: value })
+  }
+
+  return (
+    <Stack alignment="center" gap="8" className={`filters ${filtersStyles}`}>
+      <Stack alignment="center" gap="4" className={`filters ${filtersStyles}`}>
+        <Select
```
edda, line 82: I am not a fan of this filter section. You have an extra `Select` for the ready condition and pair it with our established pattern of the combo filter element where you first select the thing you filter by and then the value. Please add the filter for the ready condition inside the combo element.

apps/greenhouse/src/components/core-apps/org-admin/components/plugins/plugins/components/Toolbar.tsx
```diff
@@ -0,0 +1,133 @@
...
+            name="filterValue"
+            className="filter-value-select w-96 bg-theme-background-lvl-0"
+            value={selectedValue ? selectedValue : "Select..."}
+            onChange={handleValueChange}
+          >
+            {availableValues.map((value) => (
+              <SelectOption key={value} label={value} value={value} />
+            ))}
+          </Select>
+        </InputGroup>
```
edda, line 118: You forgot to add the filter pills so it's not possible to unselect a selected filter

apps/greenhouse/src/components/core-apps/org-admin/components/plugins/plugins/components/PluginList.tsx
```diff
@@ -30,11 +28,12 @@ const PluginList = () => {
 
   return (
     <>
-      <DataGridToolbar>
-        <ButtonRow>
+      <Stack className="mb-4">
```
edda, line 31: Please change this back to the DataGridToolbar to give the impression of one Toolbar block with buttons and filters.

edda (changes requested):

apps/greenhouse/src/components/core-apps/org-admin/components/plugins/plugins/components/Toolbar.tsx
```diff
@@ -0,0 +1,133 @@
...
+            required
+            label="Label"
+            value={selectedLabel}
+            className="filter-label-select w-64 mb-0"
+            onChange={handleLabelChange}
+          >
+            <SelectOption key="All" label="All" value="All" />
+            {labelFilters &&
+              labelFilters.map((filter) => (
+                <SelectOption key={filter?.label} label={filter?.label} value={filter?.label} />
```
edda, line 104: At least in the demo preview some of the values for this  are too long for the available width. You can solve this either with the available config props on the `Select` component (look into `truncateOptions`, this will truncate option labels that are too long). Alternatively if you don't want to truncate because it renders the value unreadable you can add a tailwind classname which forces the string to break if necessary: `break-words` on the `SelectOption`.

edda (changes requested):

apps/greenhouse/src/components/core-apps/org-admin/components/plugins/plugins/components/Toolbar.tsx
TilmanHaupt: A Filter automatically deselects, if you select another one. If there should be a combination of filters, i will need some time for another version.

apps/greenhouse/src/components/core-apps/org-admin/components/plugins/plugins/components/Toolbar.tsx
TilmanHaupt: I was refraining from supernova-style filtering, after it is wished i can change that.

uwe-mayer:
Is it a big change to implement [initialFilters](https://github.com/cloudoperators/juno/tree/b9349c7bb9df19cd6dfcbd2cf08d8fa9b269fa3b/apps/supernova#initial-filters-optional) as they are available as `prop` on supernova?

Just thinking about prefiltering the content according to `support_group` claim from the token. Would be nice to have the functionality already in place. But do not want to block or delay this PR.

WDYT @edda ?

ArtieReus:
I have to UX Remarks to fix:
- Space between message and button
Screenshot 2025-03-13 at 13 59 41
- After setting the value of a filter reset the input field
Screenshot 2025-03-13 at 13 59 47

edda (dismissed):
I'm going to approve but have two comments:
1. The message at the top of the page should have a margin-bottom so that it has some space to the "Add Plugin" button.
2. I'm not sure if the `truncateOptions` option does us any favors here with the filter labels being as long as they are. We can keep it like this for now and see what it will look like in the live system but I'm expecting this to be not really useable and we'll have to come up with a different solution, e.g. parsing the labels or going with the word-break solution instead.

uwe-mayer:
Thanks for this!

I think I am missing filtering for:
`Plugin.Spec.ClusterName`
and
`Plugin.Spec.PluginDefinition`

as in the [feature request](https://github.com/cloudoperators/juno/issues/754)

or am I missing something?

edda (approved):

andypf (changes requested):
Please address my comments

andypf (approved):

apps/greenhouse/src/components/core-apps/org-admin/components/plugins/plugins/components/Toolbar.tsx
TilmanHaupt: it is. Its a prop from the select

edda (approved):

Issue #810, opened 2025-03-11
[Bug](ui): InputGroup css is not included in prod bundle


edda:
**Description**
The `InputGroup` component has a custom css file which ensures that all the internal rounded corners of its child components are removed so that only the four external corners of the group appeared rounded. 

**Steps to Reproduce**

1. Look at any app using the `InputGroup` component, e.g. Supernova
2. Notice how the Selects and the button in the filter component in Supernova have rounded corners on all sides

**Expected Behavior**
All `scss` files should be loaded. Please fix all of them with this ticket. Here is a list of components with custom scss files:

- [ ] Button
- [ ] DateTimePicker
- [ ] FormattedText
- [ ] InputGroup
- [ ] SearchInput
- [ ] Select
- [ ] ThemeToggle
- [ ] TextInput

**Screenshots**

![Image](https://github.com/user-attachments/assets/93cc9f74-725b-4160-a235-4cd1d28c7635)

PR #814, opened 2025-03-12
fix(ui): fix InputGroup having Select elements

taymoor89:
# Summary

### What was the issue?
There was no issue in the build process it was rather an issue with the styles when there are some `Select` elements in the `InputGroup` which became obvious when I created a new story for the `InputGroup` with some `Select` elements inside just to see how it looks(see the screenshots section). 

### How I solved it?
`.juno-select` class selector was unable to find any element, as per the code it should actually be `.juno-select-toggle`. Secondly the pseudo selector `first-child` was not working when `Select` is supplied with `label` prop because it renders the label before the dropdown in the DOM and in this case dropdown is no longer the `first-child` therefore I used `first-of-type` pseudo selector. There could be a better solution from the code side but I wanted to fix it in a less intrusive way.

# Changes Made

- Fixed InputGroup styles with `Select` elements inside.

# Related Issues

- #810 

# Screenshots (if applicable)

## Before
Screenshot 2025-03-12 at 14 39 42

#### ui-components
Screenshot 2025-03-12 at 16 55 55

#### supernova
Screenshot 2025-03-12 at 16 57 42

# Checklist

# PR Manifesto


TilmanHaupt:
Nearly there - thanks a lot! But know the outer side of the outer select on the left side has no round corners. :(

taymoor89:
> Nearly there - thanks a lot! But know the outer side of the outer select on the left side has no round corners. :(

It is same as `TextInput` which also has non-rounded border, not sure if that is another bug. BTW I also like both ends of the `InputGroup` rounded just like an inputgroup with all buttons but I do not want to touch anything extra there specially when I'm not sure.

packages/ui-components/src/components/InputGroup/input-group.scss
```diff
@@ -6,8 +6,7 @@
 .juno-input-group {
   // Remove border-radius for specific elements within the input group
   .juno-button,
-  .juno-textinput,
-  .juno-select {
+  .juno-textinput {
```
edda, line 9: This can be deleted. It doesn't work, because `juno-textinput` is always wrapped in a parent element
```suggestion
   {
```

edda (changes requested):

edda (approved):
+1

TilmanHaupt (approved):

Issue #793, opened 2025-03-05
[Task](heureka): Create Service Overview Panel


hodanoori:
**Task Description**
Implement the **Service Overview Panel**, which provides an overview of issues occurring on different image versions within a service. The panel should display issue counts categorized by severity and enable navigation to service details page.

**Sub-tasks**

- [x] Fetch and display the list of images in forms of a data grid which belongs to the selected service
- [x] Display the name of Image .
- [x]  Display issue counts per severity for each image version.
- [x]  Provide navigation to the Service Details Page via the "Full Details" button for whole panel and "Show Details" button for each image version
- [x] provide the link to keppel

PR #826, opened 2025-03-18
feat(heureka): adds service overview panel

hodanoori:
# Summary

Implemented the Service Overview Panel, which provides an overview of issues occurring across different image versions within a service. The panel is displayed when clicking on a service in the Service List View. It shows issue counts categorized by severity for each image version and includes navigation options: a "Full Details" button for the entire panel and a "Show Details" button for each image version in the list.

The navigation functionality is not implemented yet, as the Service Details Page has not been developed.

# Changes Made

- Made service list items clickable.
- Displayed the Service Overview Panel when clicking on a service row in the Service List View.
- The panel header shows the selected service name and the total number of image versions.
- Added a "Full Details" button for future navigation to the Service Details Page.
- Included the image name and a link to Keppel (though the link is currently incorrect due to API limitations).
- Displayed the image version as` ... + last 8 characters` (since it's a long hash). A tooltip will display the full hash on hover. In the Service Details Page, the full version hash will be displayed later, enabling users to copy it.
- Displayed issue counts per severity. Sorting by image version and issue severity should be handled by the API but is not yet implemented.
- Used error and warning badges to highlight critical and high-severity issues for better UI design.
- Implemented pagination, which only appears when there is more than one page to display (as requested by Esther).
- Tests are added.

# Related Issues

- Issue 1: https://github.com/cloudoperators/juno/issues/793

# Screenshots (if applicable)

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist

# PR Manifesto


apps/heureka/src/components/Services/ServicesList/ServicesList.tsx
```diff
@@ -21,6 +23,34 @@ export const ServicesList = ({ filterSettings }: ServiceListProps) => {
     filterSettings,
   })
 
+  const [selectedService, setSelectedService] = useState<ServiceOverViewPanelType | null>(null)
```
taymoor89, line 26: we need to move `useFetchServices` to `Services` otherwise the list page will keep rerendering on this state change.

apps/heureka/src/components/Services/ServicesList/ServicesList.tsx
```diff
@@ -74,6 +106,7 @@ export const ServicesList = ({ filterSettings }: ServiceListProps) => {
           />
         </div>
       )}
+      {selectedService && <ServicePanel services={[selectedService]} onClose={handlePanelClose} />}
```
taymoor89, line 109: let's not pass `selectedService` as an array.

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+}
+
+type ServicePanelProps = {
+  services: ServiceOverViewPanelType[]
+  isLoading?: boolean
+  onClose: () => void
+}
+
+export const ServicePanel = ({ services = [], isLoading = false, onClose }: ServicePanelProps) => {
+  const selectedService = services[0]
```
taymoor89, line 48: let's remove that line when `services` converted to `service`

taymoor89 (commented):
Great stuff!!!

It would be great to add at least the following two tests
1. A test in `ServicesList` which makes sure that the panel is rendered when clicked on a row.
2. A test for `Panel` which makes sure that the list of component is rendered

apps/heureka/src/components/Services/ServicesList/ServicesList.test.tsx
```diff
@@ -4,24 +4,56 @@
  */
 
 import React from "react"
-import { render, screen, act } from "@testing-library/react"
+import { render, screen, act, fireEvent } from "@testing-library/react"
```
taymoor89, line 7: `fireEvent` is [deprecated](https://testing-library.com/docs/user-event/intro#differences-from-fireevent) we should use `userEvent` instead. You can see its usage in [Shell.test.tsx](https://github.com/cloudoperators/juno/blob/495b6074712d7e4e6471949af255a86482dcfa91/apps/heureka/src/components/Shell/Shell.test.tsx#L8)

apps/heureka/src/components/Services/ServicesList/ServicesList.tsx
```diff
@@ -3,23 +3,44 @@
...
+  loading,
+  error,
+  services,
+  currentPage,
+  totalNumberOfPages,
+  goToPage,
+}: ServiceListProps) => {
+  const [selectedService, setSelectedService] = useState<ServiceType | null>(null)
+
+  const handlePanelClose = () => {
```
taymoor89, line 34: we need to wrap this inside `useCallback` as well otherwise it will keep recreating on each re-render.

apps/heureka/src/mocks/fixtures/serviceImageVersions.json
```diff
@@ -0,0 +1,39 @@
...
+            "__typename": "Page"
+          }
+        ],
+        "__typename": "PageInfo"
+      },
+      "totalCount": 1,
+      "__typename": "ServiceImageVersionConnection"
+    }
+  }
+}
```
ArtieReus, line 39: missing new line here

apps/heureka/src/components/Services/Services.tsx
```diff
@@ -35,20 +36,25 @@ const getInitialFilters = (initialFilters?: InitialFilters): FilterSettings => {
...
   }
 }
-
-type Props = {
-  initialFilters?: InitialFilters
-}
-
-export const Services = ({ initialFilters }: Props) => {
-  const [filterSettings, setFilterSettings] = useState<FilterSettings>(getInitialFilters(initialFilters))
+export const Services: React.FC<{ initialFilters?: InitialFilters }> = ({ initialFilters }) => {
```
ArtieReus, line 39: I have already changed this line to remove the React.FC. How did this got in again?

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const formattedImageVersions = imageVersions.map((version) => ({
+    imageName: version.ccrn,
+    imageVersion: version.version,
+    issueCounts: version.issueCounts,
+    serviceName: service?.name,
+  }))
+
+  // Don't render anything if no service is selected
+  if (!service?.name) {
+    return null
```
ArtieReus, line 63: shouldn't here show a message instead?

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -0,0 +1,184 @@
...
+        <Stack gap="2" distribution="between" alignment="center" className="mb-2">
+          <ContentHeading
+            heading={`${capitalizeFirstLetter(service.name)} Image Versions${showTotalCount ? ` (${totalCount})` : ""}`}
+          />
+          <Button
+            variant="primary"
+            size="small"
+            onClick={() => {
+              // TODO: Navigate to service details page
+              onClose()
```
ArtieReus, line 87: why clickling here we close the panel?

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -0,0 +1,184 @@
...
+                </DataGridCell>
+                <DataGridCell>{version.issueCounts.medium || "-"}</DataGridCell>
+                <DataGridCell>{version.issueCounts.low || "-"}</DataGridCell>
+                <DataGridCell>
+                  <Button
+                    variant="primary"
+                    size="small"
+                    onClick={() => {
+                      // TODO: Navigate to service details page
+                      onClose()
```
ArtieReus, line 160: Same here. Wouldn't be better to do nothing instead of closing the panel? Or even better if there isn't yet the functionality just disable the buttons.

ArtieReus (changes requested):

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
hodanoori: This should happen, but at the moment I will disable it

apps/heureka/src/components/Services/Services.tsx
hodanoori: Maybe by rebasing it into my branch! I changes it again

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
hodanoori: I disabled the buttons for now

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
hodanoori: same here. I disabled the buttons for now

apps/heureka/src/components/Services/Services.tsx
```diff
@@ -43,12 +44,22 @@ type Props = {
 export const Services = ({ initialFilters }: Props) => {
   const [filterSettings, setFilterSettings] = useState<FilterSettings>(getInitialFilters(initialFilters))
   const { serviceFilters } = useFetchServiceFilters()
+  const { loading, error, services, currentPage, totalNumberOfPages, goToPage } = useFetchServices({
```
ArtieReus, line 47: why did you moved this to outside the ServicesList?

hodanoori: yes, because otherwise after clicking on each service in servicesList, the fetch services was also triggered. This came out from review by Taimoor

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+                  ) : (
+                    "-"
+                  )}
+                </DataGridCell>
+                <DataGridCell>{version.issueCounts.medium || "-"}</DataGridCell>
+                <DataGridCell>{version.issueCounts.low || "-"}</DataGridCell>
+                <DataGridCell>
+                  <Button
+                    disabled
+                    variant="primary"
```
edda, line 153: Primary buttons are like Highlanders, there can only be one on each page ;) 
So in this case where you map over many items and render a button for each do not use `variant="primary"` but instead just use the default variant for these cases

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
edda: If this is an intermediate state then fine, but if at time of release to users the buttons still don't do anything then it's better not to show them at all instead.

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+                        <Stack gap="1.5" alignment="center">
+                          <Icon icon="openInNew" size="16" color="jn-global-text" onClick={() => {}} />
+                          <span>Image repository</span>
+                        </Stack>
+                      </a>
+                    </Stack>
+                  </Stack>
+                </DataGridCell>
+                <DataGridCell>
+                  <Tooltip triggerEvent="hover" placement="top">
```
edda, line 129: Having looked at it some more I find that the truncated sha-hash of the image version doesn't look nice. My suggestion is to print it in full in subdued text color and smaller text size underneath the image name (where you currently already show the link to the image registry). if you have a version tag for the image you can display it here in this column

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+          <span>{capitalizeFirstLetter(service.name)} Service Overview</span>
+        </Stack>
+      }
+      opened={true}
+      onClose={onClose}
+      size="large"
+    >
+      <PanelBody>
+        <Stack gap="2" distribution="between" alignment="center" className="mb-2">
+          <ContentHeading
```
edda, line 74: You shouldn't use the `ContentHeading` component for this. You already have a heading for the Panel above. Introducing a second heading of the same semantic weight is confusing. also you're repeating the service name again. My suggestion is to add the info with the image version count directly to the PanelHeading instead.

edda (changes requested):

apps/heureka/src/components/Services/ServicePanel/ServicePanel.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import React from "react"
+import { render, screen } from "@testing-library/react"
+import { vi } from "vitest"
+import { ServicePanel } from "./ServicePanel"
+import { TestProvider } from "../../../mocks/TestProvider"
+import { ServiceType } from "../Services"
+
+// Mock the useFetchServiceImageVersions hook
+vi.mock("../useFetchServiceImageVersions", () => ({
```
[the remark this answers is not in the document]
hodanoori: we need to investigate on test using TestProvider and mock server of apollo server, as I face not finding mock data even locally using `msw`

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -54,26 +54,20 @@ export const ServicePanel = ({ service, isLoading = false, onClose }: ServicePan
...
   }))
 
   return (
     <Panel
-      heading={
-        <Stack gap="2">
-          <span>{capitalizeFirstLetter(service.name)} Service Overview</span>
-        </Stack>
-      }
+      heading={`${capitalizeFirstLetter(service.name)} Image Versions${showTotalCount ? ` (${totalCount})` : ""}`}
```
edda, line 64: If `showTotalCount` is falsy the heading will read "Servicename Image Versions" which is a bit weird. I revert my previous suggestion of adding this information to the Panel heading (sorry). I would change the Panel heading to something that really explains what we're looking at here, what you had before "Servicename Overview" is a good start imo. My new suggestion for the image count is to add it in the place where it was before but not in a `ContentHeading` component. Just add it to the `Stack` you have there and distribute it so that the button stays on the right and the text goes to the left. Then I for the info text do 

```
`${showTotalCount ? `${totalCount} image versions in service` : ""}` 
```

so that it doesn't render anything if `showTotalCount` is falsy.

apps/heureka/src/components/Services/ServicePanel/ServicePanel.tsx
```diff
@@ -109,7 +103,10 @@ export const ServicePanel = ({ service, isLoading = false, onClose }: ServicePan
               <DataGridRow key={index}>
                 <DataGridCell className="break-all overflow-hidden">
                   <Stack gap="1" direction="vertical">
-                    <span>{version.imageName}</span>
+                    <Stack gap="0.5" direction="vertical">
+                      <span>{version.imageName}</span>
+                      <span className="text-sm text-jn-global-text-subdued">{version.imageVersion}</span>
```
edda, line 108: The text color `text-jn-global-text-subdued` doesn't exist. What you want is `text-theme-light`.

Also **very important**: Never use classnames prefixed with `jn-` those are internal to juno-ui-components and it is not guaranteed that they work. instead only use the classname without the `jn-` prefix!!!

edda (changes requested):
One bug and sorry but changed my mind regarding the image version count and Panel heading.

edda (approved):
+1

ArtieReus (approved):
lgtm

PR #894, opened 2025-04-10
chore(heureka): refactors count badge component

hodanoori:
# Summary

In this PR, I removed the `SeverityCount` component and refactored the `IssueCountPerSeverity` component to also handle the single count badge functionality. Additionally, I updated the imports in the list and list item components to use the refactored version.

# Changes Made
- Removed the SeverityCount component as its functionality is now covered elsewhere
- Refactored the IssueCountPerSeverity component to include support for single count badges
- Updated imports in the list and list item components to use the refactored IssueCountPerSeverity component

# Related Issues

No issue is created for this.

# Screenshots (if applicable)

image

# Testing Instructions

1. `npm i`
2. `npm run TASK`

# Checklist

# PR Manifesto


apps/heureka/src/components/common/IssuesCountPerSeverityLevel.tsx
```diff
@@ -20,22 +20,30 @@ type IssuesCountPerSeverityLevelProps = {
...
 
-const ToolTipBadge = ({ icon, text, variant, tooltipContent, className }: ToolTipBadgeType) => (
-  <Tooltip triggerEvent="hover">
-    <TooltipTrigger>
-      <Badge icon={icon} text={text} variant={variant} className={className} />
-    </TooltipTrigger>
-    <TooltipContent>{tooltipContent}</TooltipContent>
-  </Tooltip>
-)
+export const IssueCountBadge = ({ icon, count, variant, tooltipContent, className }: IssuesCountBadgeType) =>
```
taymoor89, line 31: I'd still call it `SeverityCount` instead of `IssueCountBadge` because it shows count for a given severity level and we don't need to put `Badge` in name because that is just a representation which does not need to be part of the component name. 
Moreover I'll move this component to a separate file under `common` because this is being used across multiple components.

apps/heureka/src/components/common/IssuesCountPerSeverityLevel.tsx
```diff
@@ -20,22 +20,30 @@ type IssuesCountPerSeverityLevelProps = {
...
+  count > 0 ? (
+    <Tooltip triggerEvent="hover">
+      <TooltipTrigger>
+        <Badge
+          icon={count > 0 ? icon : undefined}
+          text={count > 0 ? count.toString() : "—"}
```
taymoor89, line 37: isn't `count` already checked in line 32

apps/heureka/src/components/common/SeverityCount.tsx
```diff
@@ -13,19 +13,22 @@ import {
...
+  count > 0 ? (
+    <Tooltip triggerEvent="hover">
+      <TooltipTrigger>
+        <Badge icon={icon} text={count.toString()} variant={variant} className={className} />
+      </TooltipTrigger>
+      <TooltipContent>{tooltipContent}</TooltipContent>
+    </Tooltip>
+  ) : (
+    <span>&mdash;</span>
+  )
```
taymoor89, line 34: I'd suggest that you add two test cases one with count and the other one with no count.

ArtieReus:
Is the dash meant to be there also?
Screenshot 2025-04-10 at 14 40 30
vs
Screenshot 2025-04-10 at 14 41 28

apps/heureka/src/components/common/IssuesCountPerSeverityLevel.tsx
```diff
@@ -50,42 +26,42 @@ export const IssuesCountPerSeverityLevel = ({ counts, displayMode = "all" }: Iss
...
       )}
-      <ToolTipBadge
+      <SeverityCount
         icon="danger"
-        text={`${counts.critical}`}
+        count={counts.critical}
         variant={counts.critical > 0 ? "danger" : "default"}
         tooltipContent="Critical Issues"
-        className={`${counts.critical > 0 ? "" : "opacity-50"}`}
+        className={counts.critical > 0 ? "" : "opacity-50"}
```
ArtieReus, line 34: If we display dashes when there is no counts do we need then `opacity-50`?

hodanoori: No, I removed it.

ArtieReus:
I will fix this issue. Aligned with @hodanoori 

Screenshot 2025-04-11 at 11 29 07

apps/heureka/src/components/common/SeverityCount.tsx
```diff
@@ -13,19 +13,27 @@ import {
...
-        <TooltipContent>{tooltipContent}</TooltipContent>
-      </Tooltip>
-    ) : (
-      <span>&mdash;</span>
-    )}
-  </>
-)
+export const SeverityCount = ({ icon, count, variant, tooltipContent, displayMode }: SeverityCountProps) => {
+  // For single mode with zero count, show a dash
+  if (displayMode === "single" && count === 0) {
```
ArtieReus, line 26: The current `displayMode` prop doesn't clearly reflect the logic of showing a dash when the count is zero — it feels a bit misleading. I’d recommend the following:

- Rename the prop to something more descriptive, like `showDashIfZero` (a boolean), which directly expresses the behavior.

- Alternatively, consider doing it outside of this component.

ArtieReus: If we decide to keep this logic within the component, we could also consider showing a tooltip when hovering over the dash. What do you think?

apps/heureka/src/components/common/IssuesCountPerSeverityLevel.tsx
```diff
@@ -4,40 +4,15 @@
...
-  <Tooltip triggerEvent="hover">
-    <TooltipTrigger>
-      <Badge icon={icon} text={text} variant={variant} className={className} />
-    </TooltipTrigger>
-    <TooltipContent>{tooltipContent}</TooltipContent>
-  </Tooltip>
-)
-
-export const IssuesCountPerSeverityLevel = ({ counts, displayMode = "all" }: IssuesCountPerSeverityLevelProps) => {
+export const IssuesCountPerSeverityLevel = ({ counts }: IssuesCountPerSeverityLevelProps) => {
```
ArtieReus, line 15: the component name is not completely correct... 
`IssuesCountPerSeverityLevel`
vs
`IssueCountsPerSeverityLevel`

apps/heureka/src/components/common/SeverityCount.tsx
hodanoori: I think displayMode is descriptive because this component either is displaying a single count badge or is displaying as part of counts for all severities. Honestly I don't think it makes sense to have a tooltip for `_` as it says that there is no important data to be shown. If you still think we should change it, let's have a call

edda: I don't think a tooltip for the dash is necessary. It should be self-explanatory

ArtieReus (approved):

taymoor89 (approved):

PR #897, opened 2025-04-15
feat(url-state-provider): rollback url-state-provider to v1.3.2

taymoor89:
# Summary

The production version of all applications currently uses `url-state-provider-v1`, which was published as a separate package (`v1.3.2` of the `url-state-provider` package) nine months ago.

The code for `url-state-provider` in the `main` branch is no longer in use. This PR essentially restores version `1.3.2`, which serves as the source of truth. Additionally, this version contains a potential bug that needs to be addressed.

**Purpose:**

This PR is a critical step towards enabling routing support in the Heureka app. After addressing the aforementioned bug, routing for Heureka will function as part of the current Greenhouse application.

# Changes Made

- Rollback [`v1.3.2`](https://github.com/cloudoperators/juno/commit/c896ff1d3eb8155f8fd99bec0f5ce0443d6676d7#diff-e2bd1f6aaaa397b1995feb6ea047bed08c5b2e90b88de5a2bd1260caf4a2cb99)
- Simply converted `.js` files to `.ts` files using `eslint-ignore` and `ts-ignore` because fighting with the types does not make sense at this point as this solution will be abandoned by the upcoming one.
- Fixed a bug where the regular expression was not ignoring the `#`ed part of the URL.

# Related Issues

- #899 

# Screenshots (if applicable)
The following screenshot clearly indicates that there is no difference in the url search params after this change which makes these changes safe to merge.

Screenshot 2025-04-15 at 17 30 30

# Testing Instructions

- Simply compare the Greenhouse URL with the Greenhouse version running on QA and there shouldn't be any change.

# Checklist

# PR Manifesto


.changeset/quiet-numbers-scream.md
```diff
@@ -0,0 +1,8 @@
+---
+"@cloudoperators/juno-url-state-provider": minor
```
ArtieReus, line 2: - @cloudoperators/juno-url-state-provider
  -  Should we still release the package or would it make more sense to make it private?

apps/heureka-legacy/package.json
```diff
@@ -50,7 +50,7 @@
     "@cloudoperators/juno-communicator": "*",
     "@cloudoperators/juno-messages-provider": "*",
     "@cloudoperators/juno-ui-components": "*",
-    "@cloudoperators/juno-url-state-provider-v1": "^1.3.2",
+    "@cloudoperators/juno-url-state-provider": "*",
```
ArtieReus, line 53: Do we want to make changes into the heureka-legacy?

apps/heureka-legacy/src/hooks/useUrlState.ts
```diff
@@ -4,8 +4,7 @@
  */
 
 import { useLayoutEffect, useEffect, useState } from "react"
-// @ts-expect-error TS(2792): Cannot find module '@cloudoperators/juno-url-state... Remove this comment to see the full error message
-import { registerConsumer } from "@cloudoperators/juno-url-state-provider-v1"
```
ArtieReus, line 8: We can leave the application as it is or? we will remove it soonish...

packages/url-state-provider/src/index.ts
```diff
@@ -3,42 +3,29 @@
...
-  title?: string
-  replace?: boolean
-}
+import LZString from "lz-string"
+import juriCutlery from "juri-cutlery"
 
-const jsonURLSerializer = superstate()
+const jsonURLSerializer = juriCutlery()
 const SEARCH_KEY = "__s"
+const regex = new RegExp(SEARCH_KEY + "=([^&#]+)")
```
ArtieReus, line 14: Do we want to add here a comment to describe why we have introduced the `#`?

packages/url-state-provider/src/index.ts
```diff
@@ -3,42 +3,29 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import LZString from "lz-string"
-import superstate from "./superstate"
+/* eslint-disable */
+// @ts-nocheck
```
ArtieReus, line 7: We might need also to describe why we switch off the ts-checks and eslint, or?

packages/url-state-provider/src/index.test.ts
```diff
@@ -3,13 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+/* eslint-disable */
```
ArtieReus, line 6: The same here.  Since we have migrated everything to ts we should then argument why we do so or?

ArtieReus (commented):
Added some comments

packages/url-state-provider/src/index.ts
barsukov: yes I agree

apps/heureka-legacy/package.json
taymoor89: I think yes, otherwise we'll not be able entirely remove the dependency `juno-url-state-provider-v1`

packages/url-state-provider/src/index.ts
taymoor89: It is mentioned in the PR description, but I can put it here as well.

apps/heureka-legacy/src/hooks/useUrlState.ts
taymoor89: Technically `juno-url-state-provider-v1` is not different that `juno-url-state-provider` after this change and because we're going to get rid of this dependency it makes sense to change here as well. In fact I think we should remove `heureka-legacy` altogether we can recover it any time from the commit history if needed.

.changeset/quiet-numbers-scream.md
taymoor89: We can make it private but I don't see a need for that because we were already publishing `url-state-provider` to `npm` and in future we need it as a public package so that the third party app developers can utilize this package.

ArtieReus: yes sure!! but now we are encoding different then before and I could break deep links to someone using the actual version...

ArtieReus (approved):

hodanoori (approved):

PR #981, opened 2025-05-28
fix(heureka): ensure that filters can be added just once

### LINKED ISSUE #974: [Bug](heureka):  possible multiplication of filters.
**Description**

You can add the same filter multiple times.  

**Expected Behavior**

Filters (key:value) can be added just once

** Tasks**
- [x] Allow Filters to just be added once
- [x] Disallow Filters to just be added if they are not possible (via deeplink) (discussion #1001 )
- [ ] Ensure that Filters which are selected are not shown in the combobox
- [ ] Check if useCallback is useful - discussion #981 
- [x] Research if the functions which altering the state should be in a extra file or at the state
- [ ] Add tests

TilmanHaupt:
# Summary

Ensures that filters can be just added once. 
Refactored useCallback

closes #974 

# Thoughts 

Even through i added my logic to the Filters component, if the State is given to multiple components from index there could be inconsistencies. I would prefer the logic near the state or in a separate file.

# Testing Instructions

1. `pnpm i`
2. `pnpm TASK`

# Checklist

# PR Manifesto

### DIFF
--- a/apps/heureka/src/components/common/Filters/index.tsx
+++ b/apps/heureka/src/components/common/Filters/index.tsx
@@ -29,18 +29,23 @@ export const Filters = ({ filters, filterSettings, onFilterChange, searchInputPl
     [filterSettings, onFilterChange]
   )
 
+  const handleFilterAdded = (filterToAdd: SelectedFilter) => {
+    // makes sure filters are just added once
+    const foundFilter = filterSettings.selectedFilters?.filter(
+      (filter) => filter.name === filterToAdd.name && filter.value === filterToAdd.value
+    )
+    if (!foundFilter || foundFilter.length == 0) {
+      onFilterChange({
+        ...filterSettings,
+        selectedFilters: filterSettings?.selectedFilters?.concat(filterToAdd),
+      })
+    }
+  }
+
   return (
     <Stack direction="vertical" gap="4" className="bg-theme-background-lvl-1 py-2 px-4 ">
       <InputGroup>
-        <FilterSelect
-          filters={filters}
-          onChange={(selectedFilter) => {
-            onFilterChange({
-              ...filterSettings,
-              selectedFilters: filterSettings?.selectedFilters?.concat(selectedFilter),
-            })
-          }}
-        />
+        <FilterSelect filters={filters} onChange={handleFilterAdded} />
         <Button
           label="Clear all"
           className="ml-4"
--- a/apps/heureka/src/components/common/Filters/index.tsx
+++ b/apps/heureka/src/components/common/Filters/index.tsx
@@ -17,17 +17,14 @@ export type FiltersProps = {
 }
 
 export const Filters = ({ filters, filterSettings, onFilterChange, searchInputPlaceholder }: FiltersProps) => {
-  const handleFilterDelete = useCallback(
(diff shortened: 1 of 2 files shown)


apps/heureka/src/components/common/Filters/index.tsx
```diff
@@ -17,30 +17,32 @@ export type FiltersProps = {
 }
 
 export const Filters = ({ filters, filterSettings, onFilterChange, searchInputPlaceholder }: FiltersProps) => {
-  const handleFilterDelete = useCallback(
-    (filterToRemove: SelectedFilter) => {
+  const handleFilterDelete = (filterToRemove: SelectedFilter) => {
```
taymoor89, line 20: `useCallback` memoize a function and recreate it only if the dependencies change, was there a specific reason to remove it? Because now this function is being recreated on each re-render.

apps/heureka/src/components/common/Filters/index.tsx
```diff
@@ -17,30 +17,32 @@ export type FiltersProps = {
...
+    onFilterChange({
+      ...filterSettings,
+      selectedFilters: filterSettings.selectedFilters?.filter(
+        (filter) => !(filter.name === filterToRemove.name && filter.value === filterToRemove.value)
+      ),
+    })
+  }
+
+  const handleFilterAdded = (filterToAdd: SelectedFilter) => {
```
taymoor89, line 29: 1. same comment about `useCallback` as above.
2. I'd call it `handleFilterAdd` instead of using past tense `handleFilterAdded` because the action is not done yet.

taymoor89 (changes requested):
please also add a test for the changed behavior to have good confidence.

apps/heureka/src/components/common/Filters/index.tsx
TilmanHaupt: Regarding the test - for me its more the task of the heureka team - I just fixed it and added some test in the other PR because i had some time over in the sprint. If I should test it i would want to change the architecture, so the functions are not at the pills anymore. I would talk about that in the sprint and somebody of the heureka team can do it.

apps/heureka/src/components/common/Filters/index.tsx
TilmanHaupt: useCallback is deprecated with 19

taymoor89:
There is change in the UX as well

### Before, after selecting the filter value the value was not staying selected in the dropdown
Screenshot 2025-05-30 at 11 58 29

### But now it keeps the value selected even after the filter is added to the list
Screenshot 2025-05-30 at 11 57 56

apps/heureka/src/components/common/Filters/index.tsx
```diff
@@ -17,30 +17,32 @@ export type FiltersProps = {
...
+    // makes sure filters are just added once
+    const foundFilter = filterSettings.selectedFilters?.filter(
+      (filter) => filter.name === filterToAdd.name && filter.value === filterToAdd.value
+    )
+    if (!foundFilter || foundFilter.length == 0) {
```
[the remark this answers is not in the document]
TilmanHaupt: Yes to ===

Issue #923, opened 2025-04-30
[Task](context): restore Firefox warning for `allow_client_cert` flag in Supernova


ArtieReus:
**Task Description**

During the Supernova migration, the warning informing Firefox users about the need to activate the  `allow_client_cert` flag was removed:
https://github.com/cloudoperators/juno/commit/a0ec74a3acbfdf3f6c7426719ffee89b8903bfd7

Firefox users without this flag enabled are currently encountering unclear CORS errors and may not understand the cause.

**Sub-tasks**

- Reintroduce the warning for Firefox users.
- Include a browser detection check to show the warning only for Firefox users.

PR #999, opened 2025-06-12
chore(supernova): restore Firefox CORS warning & re-enable ReactNode support for addMessage in messages-provider

ArtieReus:
# Summary

Restores previously `existing Firefox allow_client_cert warning` handling in Supernova alerts and extended to silences main view to allow Firefox to properly pass the client certificate during CORS requests. This functionality was accidentally removed earlier. Adds initial detection for Firefox CORS failures related to client certificates. Also restores ReactNode support, brings back the return value in addMessage(), and fixes a previously skipped test in messages-provider.

# Changes Made

- Restores Firefox-specific warning about the allow_client_cert flag in Supernova.
- Re-enables support for passing ReactNodes as message content in addMessage() in `messages-provider`.
- Restores the return value from addMessage() to return the message ID in `messages-provider`.
- Fixes a previously skipped test related to message handling in `messages-provider`.

# Related Issues

- #923

# Screenshots (if applicable)

Screenshot 2025-06-12 at 16 08 48

# Testing Instructions

1. `pnpm i`
2. `pnpx turbo dev --filter @cloudoperators/juno-app-supernova`
3. Open the App in Firefox

# Checklist

# PR Manifesto

### DIFF
--- a/apps/supernova/src/components/alerts/AlertsTab.tsx
+++ b/apps/supernova/src/components/alerts/AlertsTab.tsx
@@ -20,12 +20,13 @@ const AlertsTab = () => {
   const totalCounts = useAlertsTotalCounts()
   const updatedAt = useAlertsUpdatedAt()
   const { setAlertsData } = useAlertsActions()
-  const { addMessage } = useActions()
+  const { addMessage, resetMessages } = useActions()
 
   // Fetch alerts data
   const { data, isLoading, error } = useBoundQuery<AlertsData>("alerts")
 
   if (error) {
+    resetMessages()
     // Extra CORS warning based on instanceof
     if (error instanceof CorsNetworkError) {
       addMessage({
--- a/apps/supernova/src/components/silences/SilencesList.tsx
+++ b/apps/supernova/src/components/silences/SilencesList.tsx
@@ -39,11 +39,12 @@ const SilencesList = () => {
   const status = useSilencesStatus()
   const regEx = useSilencesRegEx()
   const { setSilences, setSilencesStatus, setSilencesRegEx } = useSilencesActions()
-  const { addMessage } = useActions()
+  const { addMessage, resetMessages } = useActions()
 
   const { data, isLoading, error } = useBoundQuery<SilencesData>("silences")
 
   if (error) {
+    resetMessages()
     // Extra CORS warning based on instanceof
     if (error instanceof CorsNetworkError) {
       addMessage({


apps/supernova/src/App.tsx
```diff
@@ -29,25 +29,25 @@ const queryClient = new QueryClient({
...
-  const fallbackRender = ({ error }: any) => {
-    return (
-      <div className="w-1/2">
-        <CodeBlock className={preErrorClasses} copy={false}>
-          {error?.message || error?.toString() || "An error occurred"}
-        </CodeBlock>
-      </div>
-    )
-  }
+const fallbackRender = ({ error }: any) => {
```
andypf, line 40: Is it really `any`? I believe you can simplify it to just `Error` or `string`

apps/supernova/src/components/alerts/AlertsTab.tsx
```diff
@@ -23,12 +23,34 @@ const AlertsTab = () => {
...
   // Fetch alerts data
   const { data, isLoading, error } = useBoundQuery<AlertsData>("alerts")
+
   if (error) {
+    // Extra CORS warning based on instanceof
+    if (error instanceof CorsNetworkError) {
+      addMessage({
+        variant: "warning",
+        text: (
+          <p>
```
andypf, line 33: It might make sense to extract this into a separate component. Just a suggestion - it's not a blocker!

apps/supernova/src/components/silences/SilencesList.tsx
```diff
@@ -43,6 +43,26 @@ const SilencesList = () => {
   const { data, isLoading, error } = useBoundQuery<SilencesData>("silences")
 
   if (error) {
+    // Extra CORS warning based on instanceof
+    if (error instanceof CorsNetworkError) {
+      addMessage({
+        variant: "warning",
+        text: (
+          <p>
```
andypf, line 51: same here

apps/supernova/src/hooks/useBoundQuery.ts
```diff
@@ -7,6 +7,32 @@ import { useQuery, UseQueryResult } from "@tanstack/react-query"
...
+    super(originalError.message)
+    this.name = "CorsNetworkError"
+    if (originalError.stack) {
+      this.stack = originalError.stack
+    }
+  }
+}
+
+function isPossibleCorsError(err: unknown): boolean {
+  const isFirefox = navigator.userAgent.includes("Firefox")
```
andypf, line 21: please move this line below the `if` statement

packages/messages-provider/src/useMessageStore.tsx
```diff
@@ -13,7 +13,7 @@ const uniqueId = (prefix: string): string => {
 
 // define the MessageProps interface
 interface MessageProps {
-  text: string
+  text?: string | ReactNode // text can be a string or a ReactNode
```
andypf, line 16: 👍

andypf (commented):
Overall, good job. I’ve left some comments with suggestions for improvement - please review and address them.

apps/supernova/src/App.tsx
ArtieReus: This is not in the scope of the pull request. I just moved some code out of the component.

apps/supernova/src/hooks/useBoundQuery.ts
```diff
@@ -7,6 +7,32 @@ import { useQuery, UseQueryResult } from "@tanstack/react-query"
...
+  constructor(originalError: Error) {
+    super(originalError.message)
+    this.name = "CorsNetworkError"
+    if (originalError.stack) {
+      this.stack = originalError.stack
+    }
+  }
+}
+
+function isPossibleCorsError(err: unknown): boolean {
```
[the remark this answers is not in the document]
ArtieReus: Since it returns a boolean, I would prefer to keep the `is` prefix. I can rename it to `isFirefoxNetworkError`. However, as you can see from the check, even if it returns true, we cannot be fully certain that it’s actually a CORS problem, Thats why I added on purpose the `possible` word.

apps/supernova/src/App.tsx
TilmanHaupt: But I am a bit surprised it is there. If I remember it right it was my job to purge anys in supernova. Or maybe just all //ts-ignore and //error? When supernova is again in focus we should def change that. And if you have time just try the `Error`

apps/supernova/src/components/alerts/AlertsTab.tsx
```diff
@@ -23,12 +24,22 @@ const AlertsTab = () => {
 
   // Fetch alerts data
   const { data, isLoading, error } = useBoundQuery<AlertsData>("alerts")
+
   if (error) {
+    // Extra CORS warning based on instanceof
+    if (error instanceof CorsNetworkError) {
+      addMessage({
+        variant: "warning",
+        text: <FirefoxCorsWarning />,
```
TilmanHaupt, line 33: If a CorsNetworkError is triggered bc of other issues or in other browsers that Error could be a bit... confusing

TilmanHaupt (dismissed):

apps/supernova/src/components/silences/SilencesList.tsx
```diff
@@ -43,6 +44,14 @@ const SilencesList = () => {
   const { data, isLoading, error } = useBoundQuery<SilencesData>("silences")
 
   if (error) {
+    // Extra CORS warning based on instanceof
+    if (error instanceof CorsNetworkError) {
+      addMessage({
+        variant: "warning",
+        text: <FirefoxCorsWarning />,
```
taymoor89, line 52: I think there are other browsers that are built on top of Firefox, IMO the messaging should not be specific to a browser rather it should be around the actual problem, as in, `CorsNetworkError` and in "how to address example" we can then mention that if you are using for example Firefox this is how this can be addressed.

apps/supernova/src/components/alerts/AlertsTab.tsx
ArtieReus: The original error still being displayed. It adds just an extra warning. That was the old functionality already. Thx!

apps/supernova/src/components/silences/SilencesList.tsx
ArtieReus: The issue just happens in Firefox because it doesn't forward the `allow_client_cert` by default. I just want to bringt back this functionality because it was accidentally removed in the passt. This is a temporary fix until we introduce BFF for greenhouse or supernova itself.

andypf (approved):

Issue #959, opened 2025-05-20
[Bug](doop): doop crashes when switching violations in detail panel after search filter is applied


ArtieReus:
**Description**

When a support group filter (e.g., "containers") is active and one item from the violations is selected a search is performed from the detail panel. Switching between violations multiple times while the panel remains open can cause the application to crash.

**Steps to Reproduce**

1. Apply the Support Group filter (e.g., "containers") in the main list.
2. Click on a violation to open the detail panel.
3. While the panel is open, perform a search in the search bar (use a string that still shows results in the main list).
4. With the panel still open, click on a different violation in the main list.
5. Repeat step 4 one or more times.

```
index-CLcsoah0.js:100 Uncaught NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at s1 (index-CLcsoah0.js:100:13527)
    at DS (index-CLcsoah0.js:183:7483)
    at HE (index-CLcsoah0.js:183:7239)
    at Wo (index-CLcsoah0.js:183:8905)
    at MS (index-CLcsoah0.js:183:9423)
    at Wo (index-CLcsoah0.js:183:9001)
    at MS (index-CLcsoah0.js:183:9124)
    at Wo (index-CLcsoah0.js:183:9001)
    at MS (index-CLcsoah0.js:183:9423)
    at Wo (index-CLcsoah0.js:183:9001)
```

**Expected Behavior**

The app should remain stable.

**Screenshots**

![Image](https://github.com/user-attachments/assets/3ec10bd2-7ca1-4ecb-b41a-32ce4c2dd806)

**Environment**

- OS: macOS
- Browser: Chrome

PR #1007, opened 2025-06-17
fix(doop): add unique id to items for stable ReactMarkdown rendering and fix inconsistent instance content

ArtieReus:
# Summary

Ensure stable rendering by adding unique ids to items for correct ReactMarkdown remounting and sanitizing instance values to prevent inconsistent DOM node structures. Fixes NotFoundError: `Failed to execute 'removeChild' on 'Node' errors caused by improper reuse of DOM nodes`

# Changes Made

**Added unique id to all items which uses ReactMarkdown**
- Each item now has a stable id (based on constraint.name, index, and a counter).
- Used as React key when rendering ReactMarkdown.
- Ensures that when the item changes, ReactMarkdown unmounts and remounts cleanly.

**Handled unsafe values in instance.cluster and instance.name**
- Some values were undefined, null, objects, or arrays.
- These now default to safe, displayable values using fallback logic (String(value), or "—").

# Related Issues

-#959

# Screenshots (if applicable)

# Testing Instructions

1. Apply the Support Group filter (e.g., "containers") in the main list.
2. Click on a violation to open the detail panel.
3. While the panel is open, perform a search in the search bar (use a string that still shows results in the main list) like ingress.
4. With the panel still open, click on a different violation in the main list.
5. Repeat step 4 one or more times.

# Checklist

# PR Manifesto

### DIFF
--- a/apps/doop/src/components/violations/ViolationDetailsList.tsx
+++ b/apps/doop/src/components/violations/ViolationDetailsList.tsx
@@ -24,9 +24,12 @@ const ViolationDetailsList = () => {
 
   const items = React.useMemo(() => {
     if (!detailsViolationGroup) return null
+    let uidCounter = 0
+
     // @ts-ignore
     return detailsViolationGroup.constraints.reduce((items: any, constraint: any) => {
       items.push({
+        id: `doc-${constraint.name}-${uidCounter++}`,
         type: "doc",
         title: constraint.name,
         severity: constraint.metadata.severity,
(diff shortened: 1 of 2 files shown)


andypf (approved):
I'm not a fan of the @ts-ignore annotations, but I understand that the entire state needs to be typesafe first.

ArtieReus:
> I'm not a fan of the `@ts-ignore` annotations, but I understand that the entire state needs to be typesafe first.

Hi! this `@ts-ignore` was already there when the typescript migration happen. I doesn't belong to this PR.

TilmanHaupt (approved):

PR #1018, opened 2025-06-23
fix(apps): fix ComboBox Reset Behavior After Selection (Regression from Select Switch)

ArtieReus:
# Summary

This PR addresses a regression introduced during the switch from Select to ComboBox, where the ComboBox would not reset after an item was selected.

**Background**

Unlike the previous Select component, ComboBox internally manages the selected value via state. Because of this, setting the same value again (e.g., "") does not trigger a rerender, preventing the underlying input from being cleared. The Select component forwarded value changes directly, resulting in consistent rerenders and expected reset behavior.

# Changes Made

- Forced a rerender of the ComboBox by using the selected value as a key, making the component uncontrolled in this specific case. This ensures the internal state is properly reset when the same value (e.g., an empty string) is set again.
- Filter button removed: The filter button has been removed in this specific UX scenario, as filters are added immediately on onChange, making the button redundant.
- ComboBox behavior has been standardized across the application in line with Juno component guidelines.

# Related Issues

- #972

# Screenshots (if applicable)

# Testing Instructions

1. `Run the applications`
2. ComboBox should be disabled
3. Choose a filter type
4. ComboBox should be enabled
5. Select a values
6. The ComboBox is reseted to empty value

# Checklist

# PR Manifesto

### DIFF
--- a/apps/doop/src/components/filters/FilterSelect.tsx
+++ b/apps/doop/src/components/filters/FilterSelect.tsx
@@ -42,6 +42,11 @@ const FilterSelect = () => {
     if (filterLabel.trim() !== "" && value.trim() !== "") {
       addFilter(filterLabel, value) // add the filter to the active filters
     }
+    // TODO: remove this after ComboBox supports resetting its value after onChange
+    // set timeout to allow ComboBox to update its value after onChange
+    setTimeout(() => {
+      setFilterValue("")
+    }, 0)
   }
 
   return (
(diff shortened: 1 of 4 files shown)


apps/doop/src/components/filters/FilterSelect.tsx
```diff
@@ -57,22 +64,17 @@ const FilterSelect = () => {
             }
           </Select>
           <ComboBox
+            key={filterValue}
```
taymoor89, line 67: [suggestion]: to add comment here so people know that this is temporary workaround until we fix combox UX.

hodanoori (approved):
lgtm

taymoor89 (approved):
Thank you for taking care of it 🙏 
Looks good to me, let's 🚢 it.

Issue #1024, opened 2025-06-25
[Task](ui): Clean up unit test console output: "The current testing environment is not configured to support act(...)"


mark-karnaukh-extern-sap:
**Task Description**

Unit tests are passing but producing console error output with the message "The current testing environment is not configured to support act(...)" when testing components with asynchronous operations or state updates. While tests are functional, these error messages create noise in test output and may mask other important warnings.
This error typically occurs when:

 - Testing components that perform async operations (API calls, timers, etc.)
 - State updates happen outside of React's awareness
 - Testing environment lacks proper React Testing Library configuration
 - Missing or incorrect Jest/testing framework setup

**Sub-tasks**

- [ ] Update Jest configuration if needed
- [ ] Ensure React Testing Library is properly configured
- [ ] Wrap async operations in act() where appropriate
- [ ] Update test utilities if necessary

**Related Issues**

**Additional Context**

PR #1034, opened 2025-06-25
test(ui): fix the rest of unit test errors

mark-karnaukh-extern-sap:
# Summary

This pull request includes fixes for the remaining unit test errors, focusing on removing unnecessary warnings from the test output. The purpose of these changes is to ensure that the test output is clean and free of extraneous warnings, improving the clarity and reliability of test results.

# Changes Made

- Fixed remaining unit test errors.
- Ensured test output does not contain unnecessary warnings.

The changes improve the quality and readability of test output without introducing new features or impacting existing functionality outside of the tests.

# Related Issues

- https://github.com/cloudoperators/juno/issues/1024

# Screenshots (if applicable)

# Testing Instructions

1. `pnpm i`
2. `pnpm TASK`

# Checklist

# PR Manifesto


packages/ui-components/src/components/Pagination/Pagination.test.tsx
```diff
@@ -432,10 +462,15 @@ describe("Pagination", () => {
     expect(screen.getByRole("textbox")).toHaveValue("")
   })
 
-  test("renders Pagination (select) with undefined totalPages prop as passed", () => {
-    act(() => {
-      render(<Pagination variant="select" totalPages={undefined} />)
-    })
+  test("renders Pagination (select) with undefined totalPages prop as passed", async () => {
+    await waitFor(() =>
```
taymoor89, line 466: We usually use `waitFor` when we do assertions not when rendering a component. I'd suggest to keep `await act` and move `await waitFor` around `expect` statement below.

packages/ui-components/src/components/ComboBox/ComboBox.test.tsx
```diff
@@ -553,7 +553,12 @@ describe("ComboBox", () => {
   })
 
   test("renders a wrapperClassName to the outer wrapping <div> element", () => {
-    render(<ComboBox wrapperClassName="my-wrapper-class" />)
+    render(
+      <AppShellProvider shadowRoot={false}>
```
taymoor89, line 557: shouldn't we use `PortalProvider` here as well just like in other tests?

packages/ui-components/src/components/Pagination/Pagination.test.tsx
taymoor89: IMO we don't even need a `waitFor`. Simply `await`ing for `act` would be sufficient, which wasn't the case before.

packages/ui-components/src/components/ComboBox/ComboBox.test.tsx
TilmanHaupt: In the end AppShellProv has the PortalProvider as well as StyleProvider in it - so all fine here.

packages/ui-components/src/components/Pagination/Pagination.test.tsx
taymoor89: @andypf true, but here I think we don't even need to use `waitFor`, wouldn't putting `await` in front of `act` work?

andypf (approved):

guoda-puidokaite:
👀 Tests failing. Can review again once fixed.

mark-karnaukh-extern-sap:
> 👀 Tests failing. Can review again once fixed.

@guoda-puidokaite where? all checks are green here.

Issue #1015, opened 2025-06-20
[Feat](ui): Make DataGridRow Hoverable


guoda-puidokaite:
### Current State

- The [DataGridRow](https://cloudoperators.github.io/juno/?path=/docs/components-datagrid-datagridrow--docs) component facilitates row rendering in a grid but currently lacks built-in hover and active state management.
- Currently, each app needs to implement their own solution, leading to inconsistency.
- Without hover feedback, user interactions might feel static and less intuitive, reducing user experience fluidity across data presentation.

<img width="1172" alt="Image" src="https://github.com/user-attachments/assets/a96769dd-985e-4845-bb8c-e8db59277459" />

### Goal

- Introduce hover and active state technologies within the DataGridRow component to enhance user interaction and navigation clarity.
- Enable boolean hover behaviour via props, supporting dynamic UI design aligned with strategic objectives for responsive interfaces.

### Task Outcome

- Implement hover and active state functionality for each DataGridRow rendering, where hover effects extend to child elements.
- Establish a prop to toggle these states, reflecting design principles as defined by team collaboration or Supernova frameworks.

---

### Estimation

- ⭐ Priority: High ?
  - _Why?_ 
  - Aligns with enhancing usability and responsiveness in data grid interactions.errors.
- ⏳ Effort: Low
  - Built upon existing grid structure; straightforward mutable transformation.
- 📈 Status: Backlog
  - Required answers to open questions

### Acceptance Criteria

- Add prop to DataGridRow to enable hover and active state.
- Implement hover/active state styling according to design below, applying:
  - cursor `cursor-pointer`
  - colour `jn-bg-theme-background-lvl-1` (same as [DataGridHeadCell](https://cloudoperators.github.io/juno/?path=/docs/components-datagrid-datagridheadcell--docs))
- Ensure all child elements respond to parent row hover states.
- Verify that the integration does not alter existing layouts or styles.
- Add unit tests for new functionality.
  - Ensure existing behaviour is unaffected, add tests if required.
- Update Storybook documentation, including usage examples.

### Technical Notes

- Apply hover logic ensuring compatibility across different screen sizes and themes.
- Consult with design specifications or external framework insights, fostering consistent UI alignment.

### Design

- Note: The mouse is hovering on row two in this example
<img width="1172" alt="Image" src="https://github.com/user-attachments/assets/9fb2153c-680e-4fec-9f56-5d7e5f8790a1" />

### Subtasks

[...]

PR #1084, opened 2025-07-10
feat(ui): make DataGridRow hoverable

guoda-puidokaite:
# Summary

Issue: https://github.com/cloudoperators/juno/issues/1015

- Added `hoverable` prop to `DataGridRow`, making the entire row in a `DataGrid` hoverable.

# Changes Made

- Fulfilled all ACs (besides announcement).
  - Ensured that the entire `DataGridRow` is hoverable, irrespective of what `children` are passed.
  - Updated Example App to use this solution.
- Removed commented-out code from `DataGridRow`.

# Checklist

# Testing

Reviewers: **Please** see `Quality Assurance` section in the [main ticket](https://github.com/cloudoperators/juno/issues/1015). 🙏

# PR Manifesto

### DIFF
--- a/apps/example/src/styles.css
+++ b/apps/example/src/styles.css
@@ -3,6 +3,5 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-@import "@cloudoperators/juno-ui-components/juno-ui-components.css";
 @import "tailwindcss";
 @import "@cloudoperators/juno-ui-components/theme.css";
--- a/apps/example/src/styles.css
+++ b/apps/example/src/styles.css
@@ -3,5 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+@import "@cloudoperators/juno-ui-components/juno-ui-components.css";
 @import "tailwindcss";
 @import "@cloudoperators/juno-ui-components/theme.css";
--- a/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
+++ b/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
@@ -8,7 +8,7 @@ import "./data-grid-row.css"
 
 export const DataGridRow = forwardRef<HTMLDivElement, DataGridRowProps>(
   ({ hoverable = false, className = "", children, ...props }, ref) => {
-    const finalClassNames = `${hoverable ? "juno-datagrid-group" : ""} juno-datagrid-row jn:contents ${className}`
+    const finalClassNames = `${hoverable ? "juno-datagrid-group" : ""} juno-datagrid-row ${className}`
 
     return (
       <div className={finalClassNames} role="row" ref={ref} {...props}>
--- a/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
+++ b/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
@@ -8,10 +8,10 @@ import "./data-grid-row.css"
 
 export const DataGridRow = forwardRef<HTMLDivElement, DataGridRowProps>(
   ({ hoverable = false, className = "", children, ...props }, ref) => {
-    const finalClassNames = `${hoverable ? "juno-datagrid-group" : ""} juno-datagrid-row ${className}`
+    const finalStyles = `${hoverable ? "juno-datagrid-group" : ""} juno-datagrid-row ${className}`
 
     return (
-      <div className={finalClassNames} role="row" ref={ref} {...props}>
+      <div className={finalStyles} role="row" ref={ref} {...props}>
         {children}
       </div>
     )
@@ -21,10 +21,10 @@ export const DataGridRow = forwardRef<HTMLDivElement, DataGridRowProps>(
 DataGridRow.displayName = "DataGridRow"
 
 export interface DataGridRowProps extends React.HTMLAttributes<HTMLDivElement> {
-  /** Determines whether the DataGridRow should apply hover styles */
+  /** Determines whether the DataGridRow should apply hover styles. */
   hoverable?: boolean
-  /** Children to render*/
-  children?: React.ReactNode
-  /** Extra styles to apply*/
+  /** Additional custom CSS class names that can be applied to the DataGridRow.  */
   className?: string
+  /** The children elements or components that will be rendered within the DataGridRow. */
+  children?: React.ReactNode
 }
--- a/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
+++ b/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
@@ -35,7 +35,7 @@ export const DataGridRow = forwardRef<HTMLDivElement, DataGridRowProps>(
     // }
 
     //  ${ selectable && isSelected ? rowSelectedStyle : '' }
-    const finalStyle = `${rowBaseStyle} ${hover ? "" : ""}`
+    const finalStyle = `${rowBaseStyle} ${hover ? rowHoverStyles : ""}`
 
     return (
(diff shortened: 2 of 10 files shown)


apps/example/src/styles.css
```diff
@@ -3,5 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+@import "@cloudoperators/juno-ui-components/juno-ui-components.css";
```
guoda-puidokaite, line 6: **Note: Importing 'Juno UI Components' Bundled Styles**

It appears that none of the other apps import this file. Instead, they redefine required classes. Should we import this file in other apps to utilise bundled styles and clean up duplications? Any context, insights?

taymoor89 (changes requested):
Nice work! 👍 
IMO test can be rewritten to confirm the actual behavior of `hoverable` prop.

packages/ui-components/src/components/DataGridRow/DataGridRow.test.tsx
```diff
@@ -18,4 +18,20 @@ describe("DataGridRow", () => {
     expect(screen.getByRole("row")).toBeInTheDocument()
     expect(screen.getByRole("row")).toHaveClass("my-custom-class")
   })
+
+  it("applies the 'juno-datagrid-group' class when hoverable is true", () => {
+    render(<DataGridRow hoverable>Content</DataGridRow>)
+    const row = screen.getByRole("row")
+
+    expect(row).toHaveClass("juno-datagrid-row")
+    expect(row).toHaveClass("juno-datagrid-group")
```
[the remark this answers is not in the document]
guoda-puidokaite: This would be the best case to test for sure, but the tests keep expecting `background-color: rgba(0, 0, 0, 0);`. I reverted to this test as I couldn't figure out the root cause. Maybe we could pair programme on it today?

packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
```diff
@@ -4,32 +4,14 @@
...
-`
-// const rowSelectedStyle = `
-// 	jn:bg-theme-datagridrow-selected
-// `
 export const DataGridRow = forwardRef<HTMLDivElement, DataGridRowProps>(
-  ({ /* selected, disabled,*/ className = "", children, /*onChange,*/ ...props }, ref) => {
-    // const dataGridContext = useDataGridContext() || {}
-    // const selectable = dataGridContext.selectable
+  ({ hoverable = false, className = "", children, ...props }, ref) => {
+    const finalStyles = `${hoverable ? "juno-datagrid-group" : ""} juno-datagrid-row ${className}`
```
edda, line 11: The name `juno-datagrid-group` doesn't suggest that this is what adds the `hoverable` behaviour. We try to give semantic names where possible. Here my suggestion is to use `juno-datagrid-row-hoverable` instead. Like this:
```suggestion
    const finalStyles = `juno-datagrid-row ${hoverable ? "juno-datagrid-row-hoverable" : ""} ${className}`
```

edda (changes requested):

packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
edda: Also please keep the classname string where it was with the `<div>`. It's not that complex yet that it requires being offloaded into a separate const and makes it easier to read.

packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
```diff
@@ -4,32 +4,14 @@
...
-    // const toggleSelected = (event) => {
-    // 	setIsSelected(!isSelected)
-    // 	onChange(event)
-    // }
-
-    //  ${ selectable && isSelected ? rowSelectedStyle : '' }
     return (
-      <div className={`juno-datagrid-row ${rowBaseStyle} ${className}`} role="row" ref={ref} {...props}>
-        {/* { selectable ? <DataGridCheckboxCell selected={selected} disabled={disabled} onChange={toggleSelected} /> : null } */}
+      <div className={finalStyles} role="row" ref={ref} {...props}>
```
edda, line 14: Please bring the class name string back here instead of offloading it into a const.

edda (changes requested):

packages/ui-components/src/components/DataGridRow/DataGridRow.test.tsx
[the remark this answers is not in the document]
edda: Also additional comment regarding this: testing for a specific background color actually tends to break more often than checking for the existence of a custom classname that we defined and won't be changed unless we want it to. However the implementaion of it (i.e. the styles that the classname applies) could potentially change more often.

packages/ui-components/src/components/DataGridRow/data-grid-row.css
```diff
@@ -0,0 +1,17 @@
...
+.juno-datagrid-row {
+  display: contents;
+}
+
+.juno-datagrid-row-hoverable:hover > * {
+  cursor: pointer;
+  background-color: var(--color-background-lvl-1);
+}
+
+.juno-datagrid-row-hoverable:active > * {
```
edda, line 15: I just noticed a change here to what we had done in the Supernova implementation. There we didn't have a style for the `:active` pseudo-selector but only for an `.active` class. This class was added to the row in the app if the row was active (i.e. selected). See [example](https://github.com/cloudoperators/juno/blob/527064c5010bbdc0ed6f36bff828e2619c32f4c1/apps/supernova/src/components/alerts/Alert.tsx#L68).

The `:active` pseudo selector only applies while the mouse button is still pressed or if using TAB navigation when pressing Return. This is good to have also but we'll need the ability to set an active state on the row. 

We should enhance the API of the component with a prop so people don't have to set undocumented classnames. We've used `isActive:boolean` in other components, so would use the same here.

edda (changes requested):

packages/ui-components/src/components/DataGridRow/data-grid-row.css
guoda-puidokaite: @edda Are you sure we'd like `.juno-datagrid-row:active`? It feels a little off. 

I think that the base row styles should remain the same with this story and that if we add interactable styling, a user should have control over disabling it. Here, they wouldn't have any.

Example of `.juno-datagrid-row:active` on click:
https://github.com/user-attachments/assets/c5f48a2e-20da-4b25-bf70-dfc5d1185a16

edda: You are right, this requires more thought. Let's not do the `:active` pseudo-selector. Just `:hover` and `.active` as discussed.

guoda-puidokaite: Thank You! I'll update ACs to change the background colour for `:hover` and `.active` and in our EPIC, add a note to consider if, when and where we should use `:active` and `cursor:pointer`. We need to think about it thoroughly, without any assumptions.

guoda-puidokaite: I'm concerned about the semantic meaning. If we break it down:

- `isHoverable` indicates a visual change on mouse over.
- `cursor: pointer` is intended for action/clicking behaviour e.g. once there's an `onClick`. 

Even though it's the most likely case that users would need hover + action... by combining them, we'd be making an assumption. E.g. What if the user wanted to allow hovering for a selection mechanism or easier information tracking, without follow up action? Example https://github.com/user-attachments/assets/0c3e9ef6-f38f-4cf2-8820-c2a9ba876df3.

We can of course do this, I would just be concerned that by combining the two, the semantic meaning changes from `isHoverable` (visual change on mouse over) to `isInteractive` (visual change on mouse over followed by clicking behaviour/action).

PR #1106, opened 2025-07-28
feat(AlertLinks) support absolute urls

trouaux:
# Summary

Some Alerts contain absolute URLs redirecting to playbooks instead of relative urls

# Changes Made

```
href={
  alert.labels.playbook.startsWith("http")
    ? alert.labels.playbook
    : `https://operations.global.cloud.sap/${alert.labels.playbook}`
}
```

# Related Issues

- Issue 1: [link to issue]

# Checklist

# PR Manifesto

### DIFF
--- a/apps/supernova/src/components/alerts/shared/AlertLinks.tsx
+++ b/apps/supernova/src/components/alerts/shared/AlertLinks.tsx
@@ -24,7 +24,11 @@ const AlertLinks = ({ alert, className }: any) => {
       {alert?.labels?.playbook && (
         <a
           className="underline"
-          href={`https://operations.global.cloud.sap/${alert?.labels?.playbook}`}
+          href={
+            alert.labels.playbook.startsWith("http")
+              ? alert.labels.playbook
+              : `https://operations.global.cloud.sap/${alert.labels.playbook}`
+          }
           target="_blank"
           rel="noopener noreferrer"
           onClick={(e) => e.stopPropagation()}
--- a/.changeset/fresh-islands-repair.md
+++ b/.changeset/fresh-islands-repair.md
@@ -1,0 +1,6 @@
+---
+"@cloudoperators/juno-app-supernova": patch
+"@cloudoperators/juno-app-greenhouse": patch
+---
+
+feat(AlertLinks) support absolute urls for playbook links.


apps/supernova/src/components/alerts/shared/AlertLinks.tsx
```diff
@@ -24,7 +24,11 @@ const AlertLinks = ({ alert, className }: any) => {
       {alert?.labels?.playbook && (
         <a
           className="underline"
-          href={`https://operations.global.cloud.sap/${alert?.labels?.playbook}`}
+          href={
+            alert.labels.playbook.startsWith("http")
```
edda, line 28: Please use optional chaining here to prevent potential runtime errors in cases when the `playbook` label doesn't exist. Like this:
```suggestion
            alert?.labels?.playbook?.startsWith("http")
```

apps/supernova/src/components/alerts/shared/AlertLinks.tsx
```diff
@@ -24,7 +24,11 @@ const AlertLinks = ({ alert, className }: any) => {
...
+              ? alert.labels.playbook
+              : `https://operations.global.cloud.sap/${alert.labels.playbook}`
```
edda, line 30: As above:
```suggestion
              : `https://operations.global.cloud.sap/${alert?.labels?.playbook}`
```

edda (changes requested):

apps/supernova/src/components/alerts/shared/AlertLinks.tsx
edda: Actually I just saw that optional chaining is used at the beginning of the block, so this part shouldn't even be reached if `playbook` or another part of the chain is undefined. So all good, no need to do this.

hodanoori (approved):

andypf (approved):
Lgtm

guoda-puidokaite (approved):

edda:
Some of our automatic checks don't work with PRs from forks. We will make some modifications that allow an admin to replace the automatic checks with manual checks in these cases. As to not block this PR I have recreated it here: #1112. Closing this.

PR #1108, opened 2025-07-29
chore(kubernetes-client): add ability to ignore ssl

andypf:
This pull request introduces two new options to the client configuration:

- **`debug`**: Enables debug logging to facilitate troubleshooting and provide more insight into client operations.  
- **`ignoreSsl`**: Allows the client to ignore SSL certificate verification errors, such as failures caused by missing or self-signed certificates during the SSL handshake. This helps prevent connection failures in development or testing environments where certificates may not be fully trusted.

### Changes  
- Added a `debug` flag to activate detailed logging output.  
- Added an `ignoreSsl` flag that, when enabled, disables SSL validation to allow connections even if certificates are invalid or missing.  
- Updated client connection logic to conditionally disable SSL verification based on the `ignoreSsl` flag.

### Motivation  
- Improve developer experience by providing detailed debug information.  
- Support use cases where SSL certificates are not properly set up (e.g., local testing, internal environments) without blocking client operations due to SSL errors.

### Security Considerations  
- The `ignoreSsl` option disables an important security check and should only be used in trusted or non-production environments.  
- Users should be made aware of the risks associated with bypassing SSL verification.

### Testing & Verification  
- Verified that enabling `debug` increases log verbosity as expected.  
- Confirmed that setting `ignoreSsl` allows connections to servers with invalid or self-signed certificates without failing.  
- Tested default behavior keeps SSL verification enabled to ensure secure connections by default.

---

Please review the implementation and provide feedback or suggestions for improvements.


hgw77 (dismissed):

hodanoori (dismissed):
I don’t like development and testing to be stopped because certificates are having issues, and I’m always a fan of detailed debug information 😊. I just have a small comment; otherwise, it’s approved from my side.

TilmanHaupt (dismissed):

PR #1153, opened 2025-09-08
feat(ui): add interaction to DataGridRow

guoda-puidokaite:
# Summary

Issue: https://github.com/cloudoperators/juno/issues/1015

- Added `isSelected` for users to "highlight" or "select" rows manually.
- Added `onClick` to `DataGridRow` which enabled interactive styles.
- If a user wants to use interactive elements inside and not have multiple events fire, they should use `event.stopPropogation()`.

# Changes Made

- Fulfilled all ACs.

# Checklist

# Testing

Reviewers: **Please** see `Quality Assurance` section in the main ticket. 🙏
All interactable elements were added to Example App and manually tested, removed for merge.

# PR Manifesto

### DIFF
--- a/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
+++ b/packages/ui-components/src/components/DataGridRow/DataGridRow.component.tsx
@@ -14,7 +14,7 @@ export const DataGridRow = forwardRef<HTMLDivElement, DataGridRowProps>(
 
     const finalStyles = `
       datagrid-row
-      ${isSelected ? "datagrid-row-active" : ""}
+      ${isSelected ? "datagrid-row-selected" : ""}
       ${onClick ? "datagrid-row-hoverable" : ""}
       ${className}
     `
--- a/packages/ui-components/src/components/DataGridRow/DataGridRow.test.tsx
+++ b/packages/ui-components/src/components/DataGridRow/DataGridRow.test.tsx
@@ -32,7 +32,7 @@ describe("DataGridRow", () => {
     const row = screen.getByRole("row")
 
     expect(row).toHaveClass("datagrid-row")
-    expect(row).toHaveClass("datagrid-row-active")
+    expect(row).toHaveClass("datagrid-row-selected")
   })
 
   it("does not apply the hoverable or active class when disabled", () => {
@@ -41,6 +41,6 @@ describe("DataGridRow", () => {
 
     expect(row).toHaveClass("datagrid-row")
     expect(row).not.toHaveClass("datagrid-row-hoverable")
-    expect(row).not.toHaveClass("datagrid-row-active")
+    expect(row).not.toHaveClass("datagrid-row-selected")
   })
 })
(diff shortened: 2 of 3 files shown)


guoda-puidokaite:
> Member

I see the failing tests, will fix them no worries. Please give some early feedback on the overall functionality, chosen components, anything missing, if concept makes sense etc.

I'll fix tests and update docs.

franzheidl:
I see, sorry if I was stating the obvious.

Since there's a lot of changed files here and since I haven't been included in the later stages of the discussion, here's my assumptions from what I see in the code and from what I remember without going through every individual file, and some early feedback based on these assumptions.

**Assumptions**
- `DataGridRow` gets `data-row-stop-propagation` attribute and `datagrid-row-hoverable`-styles when `onClick-handler is passed
- If a `DataGridCell` receives a click event, it will check whether itself or any parent have `data-row-stop-propagation` attached, and stop event propagation at that point, thus prioritising / only executing the child element's `onClick` handler
- All interactive elements get the `data-row-stop-propagation` attribute out of the box, as we can assume it's safe to always stop event propagation for these, correct?

Maybe a naive question: I am not sure I fully understand why the data-attribute to stop event propagation needs to be attached to the individual potential child elements but the event propagation needs to be actually stopped in the parent `DataGridCell`? Couldn't we let the interactive elements stop/manage their event propagation themselves (ideally have a prop control it, see my last point under *Feedback* below), the parent remaining unaware and only act on handlers once it actually receives a corresponding event (i.e. an event that was not already stopped from propagating by the child element itself)? If my view here is too simplistic, I'd absolutely love to learn.

**Feedback**
If these assumptions are correct (pls let me know if they aren't!), my quick and early feedback would be:

- Can we make this a more generic / universally usable approach that could potentially work with other components beyond `DataGridRow` that may contain interactive elements as well? If yes, I'd remove `row` from the name of the data-attribute and establish a universally usable attribute such as `data-stop-propagation` that is not semantically / by name bound to `DataGridRow` and could be used just about anywhere in the design system where we have a similar situation? Just a thought, not sure whether you have discussed anything like this with Esther before or not.
- Not sure whether `target.closest` is robust enough for all scenarios? Just a hunch from my part, from the top of my head I don't have an exact scenario at hand where it definitely wouldn't work, but wanted to share anyway
- If my third assumption is correct (we universally assign the `data-row-stop-propagation` to interactive elements) I would even more want to a point for renaming that data-attribute, as event propagation would always be stopped in the element itself, regardless whether a `DataGridRow` element is in the parent tree or not.
- If my third assumption is correct ( see previous point), I would want to make a point for not hardcoding the data-attribute, but provide a prop that controls whether a `data-stop-propagation` attribute is added to the element, or not. In the context of the assumption it's safe to always stop event propagation for affected elements, the prop could default to `true`, however I can imagine users wanting/needing the ability to keep event propagation up the DOM tree intact by explicitly setting this to `false`.

TilmanHaupt (approved):
Looks quite good. I am bit unsure about the implications other Apps. Supernova has currently a hover (grey), then its selected (lighter grey), and its hoverable again (even lighter) to show the panel will close again if you click. Here is a bit: If the function is doing something or not it will be highlighted the same way or? 

 Also, i think i found a not-the-PR-bug: Screenshot 2025-09-09 at 16 49 18

Issue #1258, opened 2025-10-29
[Task](greenhouse): always put organization name as subdomain


taymoor89:
**Task Description**

Currently, in Greenhouse, authentication requires knowing the organization name. If the organization name is missing, the user is shown an organization selector. Once an organization is selected, the information is historically persisted in the URL.

Right now, the organization name can come from **two sources**:

1. The **subdomain**, or  
2. The **first segment of the URL path** (if no subdomain is present).

This creates ambiguity and potential issues in non-development environments. The goal of this task is to standardize the handling of the organization name.

### Goals

- In **non-development environments**, the organization name should **always be in the subdomain** and **never in the URL path**.  
- In **local development**, the organization name can remain in the path because subdomains are not supported on `localhost`.

### Motivation

Standardizing the source of the organization name removes edge cases and makes authentication more predictable.

### Future Considerations

Eventually, the organization name should be stored outside the URL (e.g., in `localStorage` or cookies) to further simplify the flow.

PR #1259, opened 2025-10-29
chore(greenhouse): add org to subdomain in non-dev environments

taymoor89:
# Summary

In Greenhouse, authentication requires knowing the organization name.  
If the organization name is missing, we display an organization selector. Once the user selects an organization, we need to persist that information — historically, it has been stored in the URL.

Currently, we derive the organization name from **two possible sources**:

1. From the **subdomain**, or  
2. From the **first URL path segment** (if no subdomain is present).

This dual-source setup introduces ambiguity and issues in non-development environments.  
To simplify and make authentication more reliable, this PR ensures that:

- In **non-development environments**, the organization name is **always placed in the subdomain** and **never in the URL path**.  
- In **local development**, the organization name can still appear in the path since subdomains are not supported on `localhost`.

### Ideal solution

The ideal long-term solution would be to store the organization name outside the URL (e.g., in `localStorage` or cookies). That improvement will be addressed in a future update.

# Changes Made

- For non dev envs always put organization name as subdomain

# Related Issues

- #1258 

# Screenshots (if applicable)

# Testing Instructions

1. `pnpm i`
2. `pnpm TASK`

# Checklist

# PR Manifesto

### DIFF
--- a/apps/greenhouse/src/components/AuthProvider.tsx
+++ b/apps/greenhouse/src/components/AuthProvider.tsx
@@ -12,11 +12,10 @@ const setOrganizationToUrl = (groups: any, enableHashedRouting: boolean) => {
   if (!orgName) return
 
   let url = new URL(window.location.href)
-  const isNonDevEnv = url.host.includes("dashboard.")
 
   // In non-dev environments, set the organization as subdomain
-  if (isNonDevEnv) {
-    url.hostname = `${orgName}.dashboard.${url.hostname.replace(/^[^.]+\./, "")}`
+  if (!import.meta.env.DEV) {
+    url.hostname = `${orgName}.${url.hostname}`
     window.location.href = url.href
     return
   }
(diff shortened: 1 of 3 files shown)


apps/greenhouse/src/components/AuthProvider.tsx
```diff
@@ -7,26 +7,34 @@ import React, { createContext, useContext, useState, useMemo, useRef, useEffect
...
-      url.pathname = pathWithOrg
-    }
-    // @ts-expect-error TS(2345): Argument of type 'null' is not assignable to param... Remove this comment to see the full error message
-    window.history.replaceState(null, null, url.href)
+  const orgName = groups?.find((g: any) => g.startsWith("organization:"))?.split(":")[1]
+
+  if (!orgName) return
+
+  let url = new URL(window.location.href)
+  const isNonDevEnv = url.host.includes("dashboard.")
```
edda: This assumes that the URL of the deployed version of the app always includes the `dashboard.` substring.  We can't be sure that that is always the case. Doesn't Vite also expose the `process.env.NODE_ENV` constant or similar to be able to check what the environment is?

Issue #1268, opened 2025-11-04
[Bug](doop): initial filters aren't correctly filtering values


taymoor89:
**Description**

For each filter in `doop` we have a `type` and if the type is not know specially in case of initial filters we default it to `unknown` and after the data is fetched and that filter is part of the data then we resolve it with correct type. But that is not working after migrating to new URL structure.

**Steps to Reproduce**

1. Go to doop
2. Apply `critical` and `rotten` values for `status` filter.
3. See that the services are not filters correctly.

**Expected Behavior**
Services should be correctly filtered by the `support_group`

**Screenshots**

**Environment**

- OS: [e.g., Windows, macOS, Linux]
- Browser: [e.g., Chrome, Firefox, Safari]
- Version: [e.g., 22]

**Additional Context**

PR #1306, opened 2025-11-07
fix(doop): resolve correct filter type

taymoor89:
# Summary

In Doop, the initial filters are always typed as unknown until we fetch data and resolve them to their correct types. Previously, once the data was loaded, we updated the URL to reflect the resolved filter types.

After switching to the new URL structure, this behavior was missing — the URL no longer reflected the resolved filter types. This PR adds that functionality, ensuring that the URL correctly represents the resolved filters.

Without this fix, services were not being filtered correctly after data resolution.

# Changes Made

- Create entire filter object instead of just having `key` and `value` properties.
- Once the correct filters types are resolved reflect these in the URL state.

# Related Issues

- #1268 

# Screenshots (if applicable)

# Testing Instructions

1. `pnpm i`
2. `pnpm TASK`

# Checklist

# PR Manifesto

### DIFF
--- a/apps/doop/src/lib/urlStateUtils.test.ts
+++ b/apps/doop/src/lib/urlStateUtils.test.ts
@@ -57,26 +57,26 @@ describe("readLegacyUrlState", () => {
 describe("getFiltersForUrl", () => {
   it("should convert filter array to URL format with prefix", () => {
     const filters = [
-      { key: "status", value: "active" },
-      { key: "severity", value: "high" },
-      { key: "status", value: "pending" },
+      { key: "check:support_group", value: "active" },
+      { key: "cluster:severity", value: "high" },
+      { key: "check:support_group", value: "pending" },
     ]
 
     const result = getFiltersForUrl("f_", filters)
 
     expect(result).toEqual({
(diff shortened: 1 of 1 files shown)


apps/doop/src/lib/urlStateUtils.ts
```diff
@@ -69,22 +79,27 @@ export const convertAppStateToUrlState = (appState: any) => {
...
+    .reduce((acc: Array<Filter>, [key, value]) => {
       const filterKey = key.replace(prefix, "")
+      const [type, id] = filterKey.split(":")
+
       if (value === undefined || value === null) return acc
       // if the value is an array, add each value as a separate filter
       if (Array.isArray(value)) {
         acc = [
           ...acc,
           ...value.map((v) => ({
```
ArtieReus: Why not replicating the functionality here that we use for the initialFilters:
```
          key: `${FILTER_TYPE_UNKNOWN}:${key}`,
          id: key,
          type: FILTER_TYPE_UNKNOWN,
          label: valueToLabel(key),
          value: value,
```

PR #1376, opened 2025-12-11
fix(ui): added headlessUI in peer dependencies.

TilmanHaupt:
# Summary

- Types not exported for PopUpMenu(Item) as PopUpMenuProps is extending `<typeof MenuItem>`
- Therefore its inside the buildet `PopupMenu.component.d.ts` as:
 `import { Menu as HeadlessMenu, MenuButton, MenuItem } from '@headlessui/react';`
- HeadlessUI needed in peer dependencies or an alternative solution is required, like bundling them or add "children" and "onClick" as props

# Checklist

# PR Manifesto

### DIFF
--- a/docs/ux/0_contents.md
+++ b/docs/ux/0_contents.md
@@ -18,11 +18,11 @@
    2. [Layout Fundamentals](layout-fundamentals.md)
    3. [Navigation Fundamentals](navigation-fundamentals.md)
    4. [The AppShell](the-appshell.md)
-3. Layout And Application Structure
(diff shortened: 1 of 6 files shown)


taymoor89 (changes requested):
`@headlessui/react` is already listed as `devDependency` and bundled together with `ui-components`. Making it a peerDependency would introduce a lot of warnings if a client is not willing to supply this. I'd suggest to reconsider this approach and properly export/re-export types from the components that have propblem.

Issue #1380, opened 2025-12-12
[Task](ui): Optimization of spacing in the select component


MartinS-git:
**Task Description**

Optimization of spacing and the position of the value row in the select component to match the input field. Also, centered alignment of the icon on the right.

Too close:
<img width="401" height="74" alt="Image" src="https://github.com/user-attachments/assets/3a3e8c88-0c3d-4380-af3f-762139537ad4" />

Desired spacing: 
<img width="459" height="64" alt="Image" src="https://github.com/user-attachments/assets/8be464f4-71d6-49f5-9e24-88405ba1a517" />

PR #1381, opened 2025-12-12
fix(ui): optimize spacing in select component

MartinS-git:
# Summary

Optimization of spacing and the position of the value row in the select component to match the input field. Also, centered alignment of the icon on the right.  
=> Only the "with label" stories are affected.

# Changes Made

- Removed padding from label
- Added margin-top to value if label appears

# Related Issues + Screenshots

- Issue 1: Issue https://github.com/cloudoperators/juno/issues/1380

# Testing Instructions

1. `pnpm i`
2. `pnpm TASK`

# Checklist

### DIFF
--- a/.changeset/salty-shoes-mix.md
+++ b/.changeset/salty-shoes-mix.md
@@ -1,0 +1,5 @@
+---
+"@cloudoperators/juno-app-heureka": patch
+---
+
+implements API-Side Vulnerability Search for Service Images
--- a/apps/heureka/src/api/fetchImages.tsx
+++ b/apps/heureka/src/api/fetchImages.tsx
@@ -4,7 +4,7 @@
  */
 
 import { ApolloQueryResult } from "@apollo/client"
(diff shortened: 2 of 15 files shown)


packages/ui-components/src/components/Select/Select.component.tsx
```diff
@@ -377,7 +377,11 @@ export const Select: React.FC<SelectProps> = ({
 
                       return !hasError && !isLoading ? (
                         <>
-                          <span className={`${truncateStyles}`}>{getDisplayValue(value)}</span>
+                          <span
+                            className={`${truncateStyles} + ${label && isValueNotEmpty(label) ? "jn:mt-[0.85rem]" : ""}`}
```
edda: You don't need the `+` to concatenate the strings since you are already inside the template literal. The way you wrote it the `+` will be added as a classname, which doesn't make sense. 

```suggestion
                            className={`${truncateStyles} ${label && isValueNotEmpty(label) ? "jn:mt-[0.85rem]" : ""}`}
```

PR #1434, opened 2026-01-29
feat(docs): add Transient States And Progress

### LINKED ISSUE #1433: [Task](docs): write and add docs for transient UI states (Spinner, ProgressBar)
Write and create Juno docs how to handle transient UI states using Spinner, ProgressBar, and when to use what.

franzheidl:
Add a page to the UX docs dealing with transient UI states, loading/waiting, in-progress, Spinner and ProgressBar.

Closes #1433


docs/ux/transient-states-and-progress.md
```diff
@@ -0,0 +1,63 @@
+[← Back to Contents Overview](0_contents.md)
```
MartinS-git: I think there's a syntax error here. The content that follows is underlined.

Image

MartinS-git (dismissed):

Issue #1391, opened 2026-01-06
[Task](ui-components): decouple TabList interface from react-tabs types


taymoor89:
**Task Description**

`TabList` component interface imports `TabListProps`, `TabPanelProps`, `TabProps` from `react-tabs` package.

Issue #1390, opened 2026-01-06
[Task](ui-components): decouple DateTimePicker interface from flatpickr types


taymoor89:
**Task Description**

`DateTimePicker` interface imports `CustomLocale`, `DateLimit`, `DateOption`, `key` from `flatpickr` package.

Issue #1389, opened 2026-01-06
[Task](ui-components): decouple PopupMenu interface from @headlessui/react types


taymoor89:
**Task Description**

**Sub-tasks**

- [ ] Sub-task 1
- [ ] Sub-task 2
- [ ] Sub-task 3

**Related Issues**

**Additional Context**

Issue #1388, opened 2026-01-06
[EPIC](ui-components): decouple ui-components types from third party libraries types


taymoor89:
### Description

This epic addresses type leakage in the **ui-components** package caused by public component interfaces extending types from third-party libraries. These leaked types introduce unintended compile-time dependencies for consumers and break the abstraction boundary of the package.

#### What is the Issue?

Some UI component interfaces currently extend types from third-party libraries. While this was originally intended to improve type safety, it has resulted in strong type dependencies that leak into the public API. Consequently, consumers are required to install internal third-party packages to compile type-safe applications.

This contradicts the core design principle of the **ui-components** package: consumers should rely only on our public interfaces and remain unaware of internal implementation details, which are bundled and declared as devDependencies. Type leakage reduces encapsulation, limits future flexibility, and increases coupling between consumers and internal libraries.

---

### Objectives

- Remove third-party type dependencies from public UI component interfaces
- Ensure consumers can use ui-components without installing internal third-party packages

---

### Acceptance Criteria

- [ ] Public component interfaces no longer extend or expose third-party library types
- [ ] No breaking changes in documented component behavior or intended usage
- [ ] Type are correct in consumer environment

---

### Additional Notes

- This epic focuses on **type-level dependencies only**; runtime behavior should remain unchanged.
- Third-party libraries may still be used internally, but must be fully encapsulated.
- New component development should follow the same isolation principles established in this epic.

PR #1445, opened 2026-02-09
feat(ui): decouple interfaces from external interfaces

KirylSAP:
# Summary
Type leakage in the ui-components package caused by public component interfaces extending types from third-party libraries. These leaked types introduce unintended compile-time dependencies for consumers and break the abstraction boundary of the package.

# Changes Made

• Refactored PopupMenu public interfaces:
(PopupMenuProps, PopupMenuToggleProps, PopupMenuItemProps, PopupMenuOptionsProps) to use only our own/React types instead of types from @headlessui/react, while keeping Headless UI as an internal implementation detail.
• Introduced dedicated DateTimePicker type module (types.ts) with DateOption, DateLimit, LocaleKey, CustomLocale, and DateChangeHandler (plus backward-compatible Juno* aliases), and wired DateTimePickerProps to use these types instead of flatpickr types.
• Added internal mappers.ts for DateTimePicker to convert our public types into flatpickr types, ensuring that flatpickr remains fully encapsulated and its types are not exposed from the package.
• Updated Tab, TabList, and TabPanel components to define their own TabProps, TabListProps, and TabPanelProps instead of extending react-tabs types, keeping react-tabs usage internal.
• Adjusted packages/ui-components/src/index.ts exports so consumers import only our public DateTimePicker and PopupMenu/Tab types, with no direct dependency on third-party type definitions.

# Related Issues

- [Epic](https://github.com/cloudoperators/juno/issues/1388)
- [PopupMenu](https://github.com/cloudoperators/juno/issues/1389)
- [DatePicker](https://github.com/cloudoperators/juno/issues/1390)
- [TabList](https://github.com/cloudoperators/juno/issues/1391)

# Testing Instructions

1. `pnpm i`
2. `pnpm lint`
3. `pnpm test`

# Checklist

# PR Manifesto

### DIFF
--- a/.github/workflows/build-push-greenhouse-image.yaml
+++ b/.github/workflows/build-push-greenhouse-image.yaml
@@ -167,7 +167,7 @@ jobs:
           skip-setup-trivy: true
 
       - name: Upload Trivy scan results to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@1b168cd39490f61582a9beae412bb7057a6b2c4e # v4.31.8
+        uses: github/codeql-action/upload-sarif@19b2f06db2b6f5108140aeb04014ef02b648f789 # v4.31.11
         if: always()
         with:
           sarif_file: trivy-results.sarif
--- a/.github/workflows/build-push-supernova-image.yaml
+++ b/.github/workflows/build-push-supernova-image.yaml
@@ -186,7 +186,7 @@ jobs:
           skip-setup-trivy: true
 
       - name: Upload Trivy scan results to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@1b168cd39490f61582a9beae412bb7057a6b2c4e # v4.31.8
+        uses: github/codeql-action/upload-sarif@19b2f06db2b6f5108140aeb04014ef02b648f789 # v4.31.11
         if: always()
         with:
           sarif_file: trivy-results.sarif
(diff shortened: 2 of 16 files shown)


TilmanHaupt (dismissed):
Looks good to me.

packages/ui-components/src/components/Tab/Tab.component.tsx
```diff
@@ -62,5 +63,7 @@ export interface TabProps extends ReactTabProps {
   /** Whether the Tab is disabled */
   disabled?: boolean
   /** Add custom classNames to the Tab */
-  className?: string
+  className?: string | string[] | { [name: string]: boolean } | undefined
+  disabledClassName?: string | undefined
+  selectedClassName?: string | undefined
```
taymoor89: If possible, I'd suggest keeping `className` to `string` as before and remove `disabledClassName`, `selectedClassName` from public interface because these are internally handled.

packages/ui-components/src/components/Tab/Tab.utils.ts
```diff
@@ -0,0 +1,31 @@
...
+      if (Array.isArray(cls)) {
+        return cls
+      }
+      if (typeof cls === "object" && cls !== null) {
+        return Object.keys(cls).filter((key) => cls[key])
+      }
+      return []
+    })
+    .join(" ")
+}
```
taymoor89: If we don't change the original implementation and keep `className: string` in the component, as mentioned above, we'd not need this utility at this point.

taymoor89 (changes requested):
Thanks for taking care. Good work.
I'd suggest to change public interfaces of the components as less as possible and for standard html attributes extend with `React.HTMLElement<HTMLElement>` so we don't have to individually specify them.

packages/ui-components/src/components/Tab/Tab.component.tsx
KirylSAP: Good point! Thanks

Issue #1325, opened 2025-11-14
[Task](greenhouse): implement the plugin instance details view


ArtieReus:
**Task Description**

Implement the  plugin instance details view.

**Additional Context**

PR #1462, opened 2026-02-13
feat(greenhouse): add plugin instance detail page

taymoor89:
# Summary

This PR implements plugin instance detail page that can be accessible at `/admin/plugin-presets/$pluginPresetName/plugin-instances/$pluginInstance`.

# Changes Made

- Added plugin instance detail page
- Removed `Admin` from breadcrumb.
- Fixed conditions. Now only `Ready` condition is color coded rest of these show as default.
- Removed `Teams` page entry as we're not gonna need that. 

# Related Issues

- #1325 

# Screenshots (if applicable)

# Testing Instructions

1. `pnpm i`
2. `pnpm TASK`

# Checklist

# PR Manifesto

### DIFF
--- a/.changeset/hip-regions-deny.md
+++ b/.changeset/hip-regions-deny.md
@@ -1,0 +1,5 @@
+---
+"@cloudoperators/juno-ui-components": patch
+---
+
+Improve component descriptions and prop documentation in the editor for better developer experience
--- a/packages/ui-components/src/components/AppBody/AppBody.component.tsx
+++ b/packages/ui-components/src/components/AppBody/AppBody.component.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React from "react"
+import React, { ReactNode, HTMLAttributes } from "react"
 
 const containerStyles = `
   jn:flex
@@ -12,10 +12,15 @@ const containerStyles = `
 `
 
 /**
- * Only needed if you want to build your app's scaffold manually. In most cases it is better to use the AppShell component instead.
- * Body of the app. Treat this like the body tag of an html page.
+ * The `AppBody` component serves as the main container for the body of your application.
+ * It is specifically useful when you require manual setup of the app's layout, providing
+ * flexibility and control over the structure of the application body. For most cases,
+ * consider using the `AppShell` component which encompasses more features suitable for
+ * typical application scaffolding.
+ * @see https://cloudoperators.github.io/juno/?path=/docs/internal-appbody--docs
+ * @see {@link AppBodyProps}
  */
-export const AppBody: React.FC<AppBodyProps> = ({ className = "", children, ...props }) => {
+export const AppBody = ({ className = "", children, ...props }: AppBodyProps): ReactNode => {
   return (
     <div className={`juno-body ${containerStyles} ${className}`} {...props}>
       {children}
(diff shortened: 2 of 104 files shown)


TilmanHaupt (dismissed):

TilmanHaupt (approved):

TilmanHaupt (approved):

apps/greenhouse/src/components/admin/common/ReadinessConditions.tsx
```diff
@@ -21,10 +21,9 @@ import { createConditionAbbreviation } from "../utils"
...
   True: "success",
   False: "error",
-  Unknown: "warning",
 } as const
 
-type ConditionVariant = (typeof CONDITION_VARIANTS)[keyof typeof CONDITION_VARIANTS]
+type ConditionVariant = (typeof CONDITION_VARIANTS)[keyof typeof CONDITION_VARIANTS] | "default"
 
 type ReadinessCondition = {
   type: string
```
ArtieReus: FYI: ReadinessCondition.status is string, but you’re effectively treating it as "True" | "False" | "Unknown" | …. With the new logic, an unexpected status on Ready becomes "error"...

apps/greenhouse/src/components/admin/PluginInstanceDetail/Overview/Configuration.tsx
```diff
@@ -0,0 +1,34 @@
...
+const INJECTED_VALUES_IDENTIFIER = "global.greenhouse."
+
+export const Configuration: React.FC<ConfigurationProps> = ({ plugin }) => {
+  const optionValues = plugin.spec?.optionValues || []
+
+  return (
+    <Stack direction="vertical" gap="4">
+      <ContentHeading>Configuration</ContentHeading>
+      {optionValues.length === 0 ? (
+        <p>No configuration overrides</p>
```
ArtieReus: Should we still display here `overrides`?

apps/greenhouse/src/components/admin/PluginInstanceDetail/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Plugin } from "../types/k8sTypes"
+import { ErrorMessage } from "../common/ErrorBoundary/ErrorMessage"
+import YamlViewer from "../common/YamlViewer"
+
+const isPluginReady = (plugin: Plugin) => {
+  return plugin.status?.statusConditions?.conditions?.some((c) => c.type === "Ready" && c.status === "True")
+}
+
+const PluginInstanceDetailContent = ({ plugin }: { plugin: Plugin }) => {
+  const [selectedTabIndex, setSelectedTabIndex] = useState(0)
```
ArtieReus: PluginInstanceDetailContent keeps its own selectedTabIndex state; when the route changes to another instance, depending on whether the component remounts, the tab might remain on “YAML” unexpectedly?

edda:
> I've looked the duplication of filters in detail as in using the example `/admin/plugin-presets?f_supportGroup=containers,containers` where support group filter has `containers` value twice.
> 
> Having a filter twice is not possible from the UI but if a user messes up with the URL then they should bear the consequences and we shouldn't add more code to correct their manual changes IMO. Also it is not just about the deduplication of the filters but it applies to other search params as well and other apps share the same behavior.

Imo, as a future feature it would be nice to have an automatic deduplication in filter values but also filter keys (the other example @ArtieReus pointed out). Especially since the duplication of filter keys seems to break some of the logic with non-sensical results. We expect this to happen only if people manually manipulate the URL, and in that case, it's kind of their own fault but there might be other edge cases that cause accidental duplication (e.g. some copy-paste artifact) and it'd be nice if the apps are resilient against that. I would create a ticket for this to be picked up at some point in the future.

ArtieReus (approved):

Issue #1172, opened 2025-09-23
[Task](ui): create `DescriptionList` with child components


franzheidl:
# Create `DescriptionList` Component With Child Components

### Problem Statement
We currently do not have a standardized component/pattern to consistently display temrs and descriptions, such as metadata, etc.

### Goal
Have a consistent component in place that is to be used for displaying information metadata keys and values and similarly structured data, providing document semantics identical to the html `dl` element.

### Context
Ongoing development and enhancement of our Design System, ongoing application support.

### Implementation
Implement `DescriptionList`, `DescriptionTerm`, and `DescriptionDefinition` component that render html `<dl>`, `<dt>`, and `<dd>`elements. We should alias these as `DL`, `DT`, and `DD` accordingly, in order to provide a shorter way to write React code and allow for making shorter, better readable code, but exporting these components under their respective shortcut name, too.

When checking for types of children and deciding/implementing on layout, we should keep in mind it is possible and valid according to spec ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl)) to do the following:

- use multiple `dd` elements per `dt`
- wrap groups of a single `dt` with one or multiple `dd` elements in a `div`

**(Future) _There does not need to be a perfectly working implementation for layouting these with this first implementation, however we should be aware in order to prepare for this. Grid layout seems to be the best contender._**

#### `DescriptionList`
Renders a html `<dl>` element.

##### Props
| Prop |  Type | Default | Notes |
|------|-------|--------|-------|
| `children` | React nodes `DescriptionTerm`, `DescriptionDefinition`, `div` | `null` | allow for passign `div` as well as html allows grouping `dt` and `dl` elements in a `dl` using a `div`.  When aliasing to shorter `DT` and `DD` component names, add these as well(?) | 
| `alignTerms` | `["left", "right"]` | `"right"` | Use CSS/classes to avoid having to create a context to store purely stylistic information? | 
| `className` | `string` | `""` | | 

#### `DescriptionTerm`
Renders a html `<dt>` element.

##### Props
| Prop |  Type | Default | Notes |
|------|-------|--------|-------|
| `children` | React node | null | | 
| `className` | `string` | `""` | | 

#### `DescriptionDefinition`
Renders a html `<dd>` element.
| Prop |  Type | Default | Notes |
|------|-------|--------|-------|
| `children` | React node | null | | 
| `className` | `string` | `""` | | 

### Design Tokens

[...]

PR #1486, opened 2026-02-26
feat(ui): create DescriptionList with DescriptionTerm and DescriptionDefinition

guoda-puidokaite:
# Summary

Introduce a standardised `DescriptionList` component along with its child components, `DescriptionTerm` and `DescriptionDefinition`, to display terms and descriptions consistently across applications following semantically correct HTML elements.

- Closes https://github.com/cloudoperators/juno/issues/1172
- Previous PR https://github.com/cloudoperators/juno/pull/1329

# Changes Made

- Created `DescriptionList`, `DescriptionTerm`, and `DescriptionDefinition` Components using corresponding html elements.
- Alias Export: Provided shorter alias names (`DL`, `DT`, `DD`) for easy use in React code, promoting cleaner and more readable syntax.

# Testing Instructions

Check ACs, Figma and test in Storybook.

# Checklist

# PR Manifesto

### DIFF
--- a/packages/ui-components/src/components/DescriptionDefinition/DescriptionDefinition.test.tsx
+++ b/packages/ui-components/src/components/DescriptionDefinition/DescriptionDefinition.test.tsx
@@ -59,7 +59,7 @@ describe("DescriptionDefinition", () => {
       </DescriptionDefinition>
     )
 
-    const ddElement = screen.getByTestId("custom-id")
+    const ddElement = screen.getByRole("definition")
     expect(ddElement).toHaveAttribute("data-arbitrary-prop", "data-madeup-content")
   })
 })
--- a/packages/ui-components/src/components/DescriptionTerm/DescriptionTerm.test.tsx
+++ b/packages/ui-components/src/components/DescriptionTerm/DescriptionTerm.test.tsx
@@ -44,7 +44,7 @@ describe("DescriptionTerm", () => {
       </DescriptionTerm>
     )
 
-    const dtElement = screen.getByTestId("custom-id")
+    const dtElement = screen.getByRole("term")
     expect(dtElement).toHaveAttribute("data-arbitrary-prop", "data-madeup-content")
   })
 })


packages/ui-components/src/components/DescriptionDefinition/description-definition.css
```diff
@@ -0,0 +1,14 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
```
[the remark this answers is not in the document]
ArtieReus: it applies to all files

ArtieReus (commented):
very cool!!! Added just some thoughts

packages/ui-components/src/components/DescriptionList/description-list.css
```diff
@@ -0,0 +1,21 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.dl {
```
edda: These styles should be tailwind styles in the component itself. See above. (this will also make it easier to make the overwriteable (see below).

packages/ui-components/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  },
+}
+
+export default meta
+type Story = StoryObj<typeof DescriptionList>
+
+export const Default: Story = {
+  render: (args) => (
+    <DescriptionList {...args}>
+      <DT>Warranty</DT>
```
edda: Please add a few more dt/dd children so it's easier to understand the usecase (the typical case is multiple key/value pairs in a list)

packages/ui-components/src/components/DescriptionTerm/DescriptionTerm.component.tsx
```diff
@@ -0,0 +1,28 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { ReactNode } from "react"
+import "./description-term.css"
```
edda: replace with inline tailwind styles (see above)

packages/ui-components/src/components/DescriptionList/DescriptionList.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface DescriptionListProps {
+  /**
+   * Child components must be either DescriptionTerm or DescriptionDefinition to maintain semantic structure.
+   * Supports multiple instances to create a detailed list of terms and definitions.
+   */
+  children:
+    | ReactElement<DescriptionTermProps | DescriptionDefinitionProps>
+    | Array<ReactElement<DescriptionTermProps | DescriptionDefinitionProps>>
+    | ReactElement<"div">
```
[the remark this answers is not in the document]
franzheidl: Valid as per spec, so if we offer a component that basically just (re-)creates an html element, we should stick to the spec.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl#wrapping_name-value_groups_in_div_elements

packages/ui-components/src/components/DescriptionDefinition/DescriptionDefinition.component.tsx
```diff
@@ -0,0 +1,28 @@
...
+   * Additional class names for applying custom styles or overriding default styles on the <dd> element.
+   */
+  className?: string
+}
+
+/**
+ * Represents the definition or description in a description list, rendering as an HTML <dd> element.
+ * Pairs with DescriptionTerm to complete the term-description association, offering flexible content styling.
+ */
+export const DescriptionDefinition: React.FC<DescriptionDefinitionProps> = ({ children, className = "" }) => (
```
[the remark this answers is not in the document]
franzheidl: Interesting (and somewhat fundamental) question.

**_We generally do not check for parent validity_** (we may do in some places, but I can't think of any from the top of my head).

We may check for the **_validity of children_** in some places where we do only allow a certain type of children, and then it is not necessarily about strict html validity, but also Juno-UI-Components architectural validity, but even there I doubt we are being 100% consistent tbh.

Short answer: **In this case specifically we'd rather not do it.** 

The best we could do is display a React warning in the console OR throw an error and prevent the component from rendering. As a general rule we have decided against the latter and preserve some user responsibility. Also, users may choose to render a surrounding `dl` element themselves not using our component, and the result would be totally valid, we couldn't 100% detect that thus still emitting a warning.

packages/ui-components/src/components/DescriptionList/description-list.css
```diff
@@ -0,0 +1,21 @@
...
+  display: grid;
+  grid-template-columns: 1fr 2fr;
```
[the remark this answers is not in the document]
franzheidl: Agreed, smth in the 25% to 30% range seems a good starting point. 

We can add more complex behaviour later (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/grid-template-columns).

franzheidl (changes requested):

franzheidl (changes requested):
Basically what Esther said:

- use inline/tw-styles and remove external css
- fix default alignment of `dt`
- fix re-defining `--border-color-theme-default`
- update story wiht a few more `dt`/`dd` terms to reflect a typical use case

plus removing any changes to `Modal` from this PR.

Thanks 🙏

packages/ui-components/src/components/DescriptionList/DescriptionList.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+describe("DescriptionList", () => {
+  it("renders child DescriptionTerm and DescriptionDefinition components correctly", () => {
+    render(
+      <DescriptionList>
+        <DescriptionTerm>Term 1</DescriptionTerm>
+        <DescriptionDefinition>Definition 1</DescriptionDefinition>
+      </DescriptionList>
+    )
+    expect(screen.getByText("Term 1")).toBeInTheDocument()
+    expect(screen.getByText("Definition 1")).toBeInTheDocument()
```
franzheidl: I would definitely recommend to solidify this test, as this actually only tests whether the text passed as a child is rendered. 

We normally use roles or `testId` s to target elements, so we should use them here as well, and either test for the node type or, which I find even better, the aria-role associated with the respective elements (`term` for `<dt>`, and `definition` for `<dd>`). 

Alternatively we can test whether there are elements with the given role by trying to select them which is kind of a shortcut of the above.

In any event the test should do more of what its description claims to do IMO.

````
expect(screen.getElementByRole("term"))

````

File: packages/ui-components/src/components/DescriptionList/DescriptionList.test.tsx [hunk 9]
```diff
@@ -0,0 +1,70 @@
...
+        <DescriptionTerm>Term 5</DescriptionTerm>
+        <DescriptionDefinition>Definition 5</DescriptionDefinition>
+      </DescriptionList>
+    )
+    expect(screen.getByText("Term 4")).toBeInTheDocument()
+    expect(screen.getByText("Definition 4")).toBeInTheDocument()
+    expect(screen.getByText("Term 5")).toBeInTheDocument()
+    expect(screen.getByText("Definition 5")).toBeInTheDocument()
+  })
+})
```
franzheidl: Please add a test for arbitrary props being consumed and applied, we do this for every component (and often forget to implement as well).

PR #1504, opened 2026-03-04
feat(docs): update OverflowMenu docs single item menu policy

### LINKED ISSUE #1500: [Task](docs):  update Overflow Menu docs to reflect single item policy change
In the current documentation, we discourage users from using OverflowMenu when there is only a single item in the Menu. However recent discussions have shown this to be a viable solution in some cases, and the documentation should reflect that and give better guidance as to when to go for single item Overflow Menus.

franzheidl:
# Summary

Update ux-documentation with better guidance re garding when and when not to use single item menus in OverflowMenu.

closes #1500

### DIFF
--- a/.changeset/stale-queens-sneeze.md
+++ b/.changeset/stale-queens-sneeze.md
@@ -1,0 +1,5 @@
+---
+
+---
+
+feat(docs): update OverflowMenu docs single item menu policy


guoda-puidokaite (commented):
Some suggestions

PR #1585, opened 2026-04-08
feat(greenhouse): add Exposed Services dashboard

### LINKED ISSUE #1399: [Task](greenhouse): add exposed services list
## Task Description

Create a new admin page at `/admin/exposed-services` to list all services exposed by plugins. Uses label selector `greenhouse.sap/plugin-exposed-services=true` for efficient API-level filtering.

## Sub-tasks

- [x] Create API fetch function with label selector see the **Implementation Details** section below
- [x] Create route at `/admin/exposed-services/`
- [x] [Update] Move [ExposedServices](https://github.com/cloudoperators/juno/blob/main/apps/greenhouse/src/components/admin/PluginInstanceDetail/Overview/ExposedServices.tsx) component out of plugin instance details and change it in a way that it works for all plugins. Information about which columns to show can be found [here](https://codi.eu-nl-1.cloud.sap/kFiiCr6YQaOPZORH3hNTdQ#Exposed-Services-List-D).
- [x] [Update]Make the list filterable by support group and cluster
- [x] Add navigation menu item
- [x] [Update] Remove exposed services data grid from plugin instance detail page, and display space separated list of expose services names as hyperlink in basic information sections.
- [x] Test with real data

## Implementation Details
In order to fetch all services exposed by all plugins following method can be used.

```
/*
 * SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Juno contributors
 */

import { Plugin } from "../../types/k8sTypes"

export const FETCH_EXPOSED_SERVICES_CACHE_KEY = "exposedServices"

export const fetchExposedServices = async ({
  apiClient,
  namespace,
}: {
  apiClient: any
  namespace: string
}): Promise<Plugin[]> => {
  // Filter only by plugins that have exposed services
  const labelSelector = "greenhouse.sap/plugin-exposed-services=true"

  const response = await apiClient.get(`/apis/greenhouse.sap/v1alpha1/namespaces/${namespace}/plugins`, {
    params: {
      labelSelector,
    },
  })
  return Array.isArray(response?.items) ? response.items : []
}
```

## Related Issues

- https://github.com/cloudoperators/greenhouse/issues/1555

## Additional Context

Please follow established Greenhouse admin patterns with React Query, DataGrid components, and proper error handling.

guoda-puidokaite:
# Summary

Add a list of exposed services for all clusters to `Greenhouse` under `Exposed Services` navigation tab

**Exposed Services**

Filters by `supportGroup` as default. This has all filters cleared.

**EXPOSED SERVICES LIST**

Screenshot 2026-04-19 at 23 49 46

---

**PLUGIN INSTANCE DETAIL PAGE**

**Before**

Screenshot 2026-04-19 at 23 15 05

**After**

Screenshot 2026-04-19 at 23 16 00

# Changes Made

- Added a list of exposed services, see ACs
- Removed list of exposed services from Plugin Instance detail page and added them as a list in `Details`
- Updated `PluginPresets` page to use `DataGridToolbar`

**NOTE** Will raise a separate issues for:
- Selected filters should be removed from the dropdown cross pages
- Cross page linking e.g. when you press on a cluster in `Exposed Services` page, navigating to its detail page

# Related Issues

- Closes https://github.com/cloudoperators/juno/issues/1399

# Screenshots (if applicable)

# Testing Instructions

1. Review ACs
2. Test Greenhouse preview link

# Checklist

# PR Manifesto


ArtieReus (commented):
Short question... I see that not just the link is clickable instead the whole cell. Is this intentionally? and if yes is this agreed? we don't have this pattern implemented. From my point of view it should be just the link and not the whole cell.

Image

ArtieReus (commented):
Shouldn't the selected filter be removed from the available filters in the dropdown?

Image

apps/greenhouse/src/components/admin/api/exposed-services/fetchExposedServicesFilters.ts
```diff
@@ -0,0 +1,59 @@
...
+const getSupportGroupValues = (services: PluginPreset[]) =>
+  Array.from(
+    new Set(
+      services.map((service) => {
+        return service.metadata?.labels?.[SUPPORT_GROUP_LABEL]
+      })
+    )
+  ).filter((value): value is string => !!value)
+
+const extractilters = (exposedServices: PluginPreset[]) => {
```
ArtieReus: why are the filters hardcoded and not dynamic collected?

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+        return (
+          <DataGridRow key={`${pluginName}-${serviceUrl}`}>
+            {/* Name */}
+            <DataGridCell>
+              {serviceUrl ? (
+                <a href={serviceUrl} target="_blank" rel="noopener noreferrer" className="cursor-pointer">
+                  <Stack gap="2" alignment="center">
+                    {serviceData.name || ""}
+                    {/* Remove onclick - components bug */}
```
ArtieReus: what does coment means?

apps/greenhouse/src/components/admin/api/exposed-services/fetchExposedServicesFilters.ts
ArtieReus: And I see also a typo on the function name `extractilters`.

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+        const serviceUrl = exposedServices ? Object.keys(exposedServices)[0] : ""
+        const serviceData = exposedServices ? Object.values(exposedServices)[0] : { name: "", namespace: "" }
+
+        return (
+          <DataGridRow key={`${pluginName}-${serviceUrl}`}>
+            {/* Name */}
+            <DataGridCell>
+              {serviceUrl ? (
+                <a href={serviceUrl} target="_blank" rel="noopener noreferrer" className="cursor-pointer">
```
ArtieReus: Missing aria-labels...

ArtieReus (commented):
Icon seems to float away from string

Image

guoda-puidokaite:
> Shouldn't the selected filter be removed from the available filters in the dropdown?
> 
> Image

**(Future Improvement, Existing Functionality)**
I think it would be better... I already noted this as a general improvement across views (_updated PR description_). However, it's something we'll need to to discuss. This is how it works across views right now e.g. check `PluginPresets`. Instead of this, in the code we currently keep all options and check "if filter isn't already applied, apply it".

guoda-puidokaite:
> Short question... I see that not just the link is clickable instead the whole cell. Is this intentionally? and if yes is this agreed? we don't have this pattern implemented. From my point of view it should be just the link and not the whole cell.
> 
> Image

**(Overwrote Default Span Behaviour)**
Good catch, needed to override default `flex` behaviour which spanned the link across the entire cell. But I think this is self explanatory, it's a link that should be clickable, not the cell. Thank You!

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
guoda-puidokaite: Issue fixed, comment removed.

apps/greenhouse/src/components/admin/api/exposed-services/fetchExposedServicesFilters.ts
guoda-puidokaite: Seems like unnecessary overhead. I went with the example in `PluginPresets` as it makes things simple and consistent with just two filters. If we had dynamic filters or many more (something more complex), we could do it that way. If something changed, we'd adapt in all views... but don't think we need to design for it now. Let me know if you see a strong argument for it though.

edda:
> Shouldn't the selected filter be removed from the available filters in the dropdown?
> 

Yes, it should.

apps/greenhouse/src/components/admin/PluginInstanceDetail/Overview/Details.tsx
```diff
@@ -12,53 +12,78 @@ import {
...
-        </DataGridRow>
-        <DataGridRow>
-          <DataGridHeadCell nowrap>Cluster</DataGridHeadCell>
-          <DataGridCell>{plugin.spec?.clusterName}</DataGridCell>
-        </DataGridRow>
-        {plugin.metadata?.labels && Object.keys(plugin.metadata.labels).length > 0 && (
+export const Details: React.FC<DetailsProps> = ({ plugin }) => {
+  const exposedServices = plugin.status?.exposedServices || {}
+  const exposedServicesLinks = Object.entries(exposedServices).map(([url, service], index) => (
+    <span key={url}>
```
edda: You should probably have a fallback for the `key` here if it's a possibility that `url` is undefined or null. Otherwise you might run into the issue that multiple entries get the same key if multiple have undefined urls.

apps/greenhouse/src/components/admin/PluginInstanceDetail/Overview/Details.tsx
```diff
@@ -12,53 +12,78 @@ import {
...
+                  {Object.entries(plugin.metadata.labels).map(([key, value]) => (
+                    <Pill key={key} pillKey={key} pillValue={value} />
+                  ))}
+                </Stack>
+              </DataGridCell>
+            </DataGridRow>
+          )}
+          <DataGridRow>
+            <DataGridHeadCell nowrap>Exposed Services</DataGridHeadCell>
+            <DataGridCell>{exposedServicesLinks.length > 0 ? exposedServicesLinks : ""}</DataGridCell>
```
edda: The fallback shouldn't be an empty string. We try to be explicit with empty states so the users know it's not some sort of error that prevented the information from rendering. Our established pattern here is to use `--`
```suggestion
            <DataGridCell>{exposedServicesLinks.length > 0 ? exposedServicesLinks : "--"}</DataGridCell>
```

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
```diff
@@ -0,0 +1,86 @@
+/*
+ * SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Juno contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from "react"
+import { useQuery } from "@tanstack/react-query"
+import { useRouteContext, useSearch } from "@tanstack/react-router"
+import { DataGridRow, DataGridCell, Icon, Stack } from "@cloudoperators/juno-ui-components"
```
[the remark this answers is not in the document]
guoda-puidokaite: Done

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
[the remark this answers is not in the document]
guoda-puidokaite: Old

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/index.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    expect(screen.getByText("Cluster")).toBeInTheDocument()
+    expect(screen.getByText("Plugin")).toBeInTheDocument()
+    expect(screen.getByText("Support Group")).toBeInTheDocument()
+
+    // Check for data - exposed services rendered
+    expect(screen.getByText("service1")).toBeInTheDocument()
+
+    // Check for data - cluster rendered
+    expect(screen.getByText("demo")).toBeInTheDocument()
+  })
```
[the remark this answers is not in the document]
guoda-puidokaite: Old

apps/greenhouse/src/components/admin/api/exposed-services/fetchExposedServicesFilters.ts
```diff
@@ -0,0 +1,59 @@
...
+  return [
+    {
```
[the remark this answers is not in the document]
guoda-puidokaite: Old

apps/greenhouse/src/components/admin/PluginInstanceDetail/Overview/Details.tsx
```diff
@@ -12,53 +12,78 @@ import {
...
+export const Details: React.FC<DetailsProps> = ({ plugin }) => {
+  const exposedServices = plugin.status?.exposedServices || {}
+  const exposedServicesLinks = Object.entries(exposedServices).map(([url, service], index) => (
+    <span key={url}>
+      {url ? (
+        <a href={url} target="_blank" rel="noopener noreferrer">
+          <Stack gap="2">
+            {service.name}
+            {/* Remove onclick - components bug */}
+            <Icon size="18" color="jn-global-text" icon="openInNew" onClick={() => {}} />
```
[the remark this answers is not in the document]
guoda-puidokaite: Old

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
```diff
@@ -35,43 +35,49 @@ export const DataRows = ({ colSpan }: DataRowsProps) => {
...
-            </DataGridCell>
-            {/* Cluster */}
-            <DataGridCell>{clusterName}</DataGridCell>
-            {/* Plugin */}
-            <DataGridCell>{pluginName}</DataGridCell>
-            {/* Support Group */}
-            <DataGridCell>{supportGroup}</DataGridCell>
+          <DataGridRow key={`${pluginName}`}>
+            {Object.entries(exposedServices).map(([url, serviceData]) => (
+              <React.Fragment key={url}>
```
edda: The comment I made somewhere else regarding `url` is also valid here. If `url` can be undefined/null (as is suggested by the fallback a few lines later) then you should add some fallback to this because otherwise you might get errors from React that the keys aren't unique

guoda-puidokaite: Ooops, you're right. There were two remaining cases of this. Fixed.

apps/greenhouse/src/components/admin/constants.ts
```diff
@@ -6,8 +6,11 @@
 export const SELECTED_FILTER_PREFIX = "f_"
 
 export const SUPPORT_GROUP_LABEL = "greenhouse.sap/owned-by"
+export const CLUSTER_LABEL = "greenhouse.sap/cluster"
+export const EXPOSED_SERVICES_LABEL = "greenhouse.sap/plugin-exposed-services=true"
```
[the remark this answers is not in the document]
guoda-puidokaite: Won't do

apps/greenhouse/src/components/admin/types/k8sTypes.ts
```diff
@@ -7,5 +7,6 @@ import type { components } from "./schema"
 
 export type PluginPreset = components["schemas"]["PluginPreset"]
 export type Plugin = components["schemas"]["Plugin"]
+export type ExposedServices = components["schemas"]["ExposedServices"]
```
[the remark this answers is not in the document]
guoda-puidokaite: Won't do, changed

apps/greenhouse/src/components/admin/ExposedServices/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <ContentHeading>Exposed Services Overview</ContentHeading>
+        <p>Monitor exposed services for all plugins</p>
+      </Container>
+      <Container px={false} py>
+        <ExposedServicesFilters />
+        <DataGridToolbar alignRight={false}>
+          <Stack>
+            <div>
+              <span className="text-theme-default pr-2">{`${total} exposed services`}</span>
+            </div>
```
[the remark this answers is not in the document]
guoda-puidokaite: Changed code significantly to flatten services, not an issue

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+interface DataRowsProps {
+  colSpan: number
+}
+
+export const DataRows = ({ colSpan }: DataRowsProps) => {
+  const { apiClient, user } = useRouteContext({ from: "/admin/exposed-services" })
+  const search = useSearch({ from: "/admin/exposed-services" })
+  const filterSettings = extractFilterSettingsFromSearchParams(search)
+
```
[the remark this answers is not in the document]
guoda-puidokaite: This is happening cross page, will fix in another PR.

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  if (!flattenedExposedServices || flattenedExposedServices.length === 0) {
+    return <EmptyDataGridRow colSpan={colSpan}>No exposed services found.</EmptyDataGridRow>
+  }
+
+  return (
+    <>
+      {flattenedExposedServices.map((service, index) => (
+        <DataGridRow key={`${service.serviceName}-${index}`}>
```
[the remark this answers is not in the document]
guoda-puidokaite: Not likely scenario, need unique key in this case

apps/greenhouse/src/components/admin/PluginInstanceDetail/Overview/Details.tsx
```diff
@@ -12,53 +12,72 @@ import {
...
-        <DataGridRow>
-          <DataGridHeadCell nowrap>Cluster</DataGridHeadCell>
-          <DataGridCell>{plugin.spec?.clusterName}</DataGridCell>
-        </DataGridRow>
-        {plugin.metadata?.labels && Object.keys(plugin.metadata.labels).length > 0 && (
+export const Details: React.FC<DetailsProps> = ({ plugin }) => {
+  const exposedServices = plugin.status?.exposedServices || {}
+  const exposedServicesEntries = Object.entries(exposedServices)
+  const exposedServicesLinks = exposedServicesEntries.map(([url, service], index) => (
+    <span key={`${url}-${service.name}-${index}`}>
```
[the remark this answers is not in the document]
guoda-puidokaite: Not likely scenario, need unique key in this case

apps/greenhouse/src/components/admin/common/ExternalLink.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return safeUrl ? (
+    <a
+      href={safeUrl}
+      target="_blank"
+      rel="noopener noreferrer"
+      aria-label={`Open ${displayName} in a new tab`}
+      className="cursor-pointer"
+    >
+      <div className="inline-block mr-2">{displayName}</div>
```
edda: Instead of settting `inline-block` on both siblings use a flex layout on the parent with `align-items: center`. This will ensure perfect alignment between text and icon. Always use this method when aligning text with icons.

ArtieReus: I don't know if @edda improvement fixes displaying the icon in a new row:
Screenshot 2026-04-27 at 13 50 15

ArtieReus (commented):
As I mentioned in today’s meeting, there’s an issue with the ComboBox search: after selecting an item, the internal search state is not reset. As a result, users cannot see the full list of items for a filter type—even when switching back and forth between options. See video:
 
https://github.com/user-attachments/assets/879fe1df-31c7-4ec1-9919-580d9011233d

You can fix it remounting the ComboBox with a dynamic key on item selected:
https://github.com/cloudoperators/juno/pull/1619/changes#diff-5b0ce3c24c269122c31734a2ed542d7a33c0820ea6e87e5864e516abd9f5b57aR38

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesFilters.tsx
```diff
@@ -0,0 +1,127 @@
...
+        />
+        <SearchInput
+          placeholder={`search term for exposed service name`}
+          className="w-96 ml-auto"
+          data-testid="searchbar"
+          value={filterSettings.searchTerm}
+          onSearch={(searchTerm) => {
+            updateFilters({
+              ...filterSettings,
+              searchTerm,
```
ArtieReus, line 98: Just a minor improvement. We could trim here the string just in case someone types a white space before and after accidentally.

apps/greenhouse/src/components/admin/common/ExternalLink.tsx
guoda-puidokaite: @edda Please see https://github.com/cloudoperators/juno/pull/1585#issuecomment-4281212203. The component already uses `flex`, I implemented the solution you propose here initially... But I need only the `children` to be clickable and for it to not `span` the rest of the cell. Also https://github.com/cloudoperators/juno/pull/1585#pullrequestreview-4139832100.

guoda-puidokaite: > I don't know if @edda improvement fixes displaying the icon in a new row: Screenshot 2026-04-27 at 13 50 15

I don't see this as a big issue, nicer than https://github.com/cloudoperators/juno/pull/1585#pullrequestreview-4139832100 and https://github.com/cloudoperators/juno/pull/1585#issuecomment-4281212203 that I already addressed. I will take note to fix it in another PR though... if okay with you, @ArtieReus? 🙈

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesFilters.tsx
guoda-puidokaite: If we do this, should keep it consistent... So will take note as improvement across all tabs.

guoda-puidokaite:
> As I mentioned in today’s meeting, there’s an issue with the ComboBox search: after selecting an item, the internal search state is not reset. As a result, users cannot see the full list of items for a filter type—even when switching back and forth between options. See video:
> 
>  Screen.Recording.2026-04-27.at.14.07.43.mov 
> You can fix it remounting the ComboBox with a dynamic key on item selected: https://github.com/cloudoperators/juno/pull/1619/changes#diff-5b0ce3c24c269122c31734a2ed542d7a33c0820ea6e87e5864e516abd9f5b57aR38

I already addressed the next steps for this issue https://github.com/cloudoperators/juno/pull/1585#issuecomment-4281008036. I want to fix cross-tab bugs and inconsistencies together, keeping this PR on-topic of new functionality.

guoda-puidokaite:
Thank You so much @edda and @ArtieReus for the thorough review. 🚀 I left comments on all the recent feedback. All issues that are cross-tab and not related to new functionality will be addressed in the next PR, keeping this one clean with only new functionality (besides one small change in `PagePresets` to use `DataGridToolbar`).

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesFilters.tsx
ArtieReus: Please link the ticket with all those improvements being done in a new ticket so we can follow from this PR. Thanks!!

ArtieReus (dismissed):
Gooooood work! 🎉
just please remember to link the issue with the follow up improvements. Thanks!!

apps/greenhouse/src/components/admin/common/ExternalLink.tsx
edda: I might not have explained in enough detail. What I meant was giving the `<a>` a flexbox layout with align items center and removing the `inline-block` from the children

ArtieReus (dismissed):

apps/greenhouse/src/components/admin/common/ExternalLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return safeUrl ? (
+    <a
+      href={safeUrl}
+      target="_blank"
+      rel="noopener noreferrer"
+      aria-label={`Open ${displayName} in a new tab`}
+      className="cursor-pointer"
+    >
+      <Stack gap="2" alignment="center">
+        <div className="mr-2">{displayName}</div>
```
edda: with the `gap` on the Stack you won't need the margin here anymore.
```suggestion
        <div>{displayName}</div>
```

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesDataGrid/DataRows/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <DataGridCell className="inline-block">
```
edda: The `inline-block` isn't needed anymore after going back to Stack.
```suggestion
          <DataGridCell>
```

apps/greenhouse/src/components/admin/ExposedServices/ExposedServicesFilters.tsx
```diff
@@ -0,0 +1,127 @@
...
+              if (!filterExists) {
+                updateFilters({
+                  ...filterSettings,
+                  selectedFilters: [...(filterSettings.selectedFilters || []), selectedFilter],
+                })
+              }
+            }}
+          />
+        </InputGroup>
+        <Button
```
edda: Since I commented this in the Clusters PR, here as well to make it consistent: 
Please use the pattern of xs button in the pill row directly behind the pills, only visible if at least one filter is active

edda (approved):

ArtieReus (approved):

PR #1628, opened 2026-04-23
feat(ui): add Divider component

### LINKED ISSUE #1612: [Task](ui): create `Divider` component
Add a general-purpose Divider component that by default renders a 1px horizontal line with configurable vertical spacing and optional color override for separating sections of content in layouts, panels, forms, etc.

### Design Considerations (Resolved)
- Named `Divider` — horizontal by default; a future vertical variant would use an `orientation`or `direction` prop
- Two-element structure to avoid margin collapse: outer `<div>` carries spacing via **CSS padding** (to avoid margin collapsing issues), inner `<hr>` is the visible 1px line divider, the `hr` element gives us the semantic `separator` role for free.
- `spacing` prop name — abstracts away the CSS mechanism, consistent with design-system vocabulary and what we do in `Stack`
- `color` accepts any tailwind border-color token and applies it on the inner element — `className` targets the outer wrapper so callers have no other way to restyle the line without reaching through the DOM. 
- Spreads remaining `HTMLAttributes<HTMLDivElement>` onto the outer wrapper. Supports `forwardRef`.

### Props

| Name | Type | Default | Description |
|-------|------|--------|-------------|
| `spacing`| Tailwind spacing token (same shape as Stack's gap) | `1`| Vertical space around the line |
| `color`| `string`| `undefined` | Overrides the line color. Accepts any valid tailwind border color token. Defaults to theme default border color |
| `className`| `string` |`""`| Applied to the outer wrapper |

### Acceptance Criteria
- [x]  Renders a full-width 1px line using the theme default border color by default
- [x]  `color` prop overrides the line color; accepts any valid CSS color value
- [x]  Default spacing="1" 
- [x]  No margin collapse regardless of surrounding elements
- [x]  `className` and all HTML `div` props apply to the outer wrapper
- [x]  Storybook story showing default and custom color and custom spacing variant
- [x]  Unit tests covering default render, color, and spacing variations
- [x]  Exported from the package index

franzheidl:
Adds a generic `Divider`component with adjustable color and spacing.

For the spacing logic also see my comment in the original issue #1612 .

Closes #1612


.changeset/large-mails-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudoperators/juno-ui-components": patch
```
edda: This should be a minor version increase, not patch.
```suggestion
"@cloudoperators/juno-ui-components": minor
```

edda (approved):

MartinS-git (approved):

PR #1644, opened 2026-04-30
feat(ui): add extra-small size to Button component

### LINKED ISSUE #1638: [Task](ui): Add extra-small size to Button component
**Description**

Add a new `extra-small` size variant to the `Button` component. This size is already defined in the Figma design system and is particularly useful in space-constrained contexts like the DataGrid toolbar, where actions like "Clear all" should have less visual weight and footprint.

**Design Reference**

Button Extra Small (all variants available: Primary, Default, Primary_Danger, Subdued):

Button Extra Small – Figma

Usage context – DataGrid Toolbar with filter pills and "Remove filters" button (extra-small):

DataGrid Toolbar – Figma

---

## Phase 1: Button component

Add `extra-small` as a new size option to the Button component.

- [ ] Add `"extra-small"` to the `ButtonSize` type
- [ ] Define Tailwind classes for extra-small typography and padding (all variants)
- [ ] Update `getButtonPadding`, `iconClasses`, and Spinner size logic to handle the new size
- [ ] Add extra-small stories to `Button.stories.tsx`
- [ ] Update existing tests or add new tests for the extra-small size
- [ ] Verify visual alignment with Figma

## Phase 2: DataGridToolbar usage

Use the new extra-small Button in the DataGridToolbar story as a real-world usage example.

- [ ] Update `DataGridToolbar` story "Complex Custom Layout": change the "Clear all" button to use `size="extra-small"`
- [ ] Update story description to mention the extra-small button usage

MartinS-git:
* Adds `xs` as a new `ButtonSize` variant
* Defines Tailwind classes for xs typography, padding, and icon size
* Extracts `getButtonBase()` and `iconSize()` as helper functions (replacing inline ternaries)
* Adds `XS`, `XSWithIcon`, and `XSInProgress` stories
* Adds unit tests for the new size
* Updates `DataGridToolbar` story: "Clear Filters" button now uses `size="xs"`
* Note: Spinner default margin adjustment will be addressed in a separate PR

Closes [#1638](https://github.com/cloudoperators/juno/issues/1638)

### DIFF
--- a/packages/ui-components/src/components/Button/Button.component.tsx
+++ b/packages/ui-components/src/components/Button/Button.component.tsx
@@ -61,7 +61,7 @@ const btnSmallSubduedPadding = `
 
 const btnExtraSmallDefaultPadding = `
   jn:py-1
-  jn:px-1
+  jn:px-1.5
 `
 
 // default size padding
@@ -229,9 +229,11 @@ export const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonPr
     const titleValue = title || label || ""
 
     const buttonIcon = progress ? (
-      <span className={`${label || children ? iconClasses(size) : ""}`}>
-        <Spinner size={`${iconSize(size)}`} color={spinnerColorClass(variant)} className={`jn:!mr-0`} />
-      </span>
+      <Spinner
+        size={`${iconSize(size)}`}
+        color={spinnerColorClass(variant)}
+        marginRight={`${label || children ? iconClasses(size) : ""}`}
+      />
     ) : icon ? (
       <Icon
         icon={icon}
--- a/packages/ui-components/src/components/Button/Button.component.tsx
+++ b/packages/ui-components/src/components/Button/Button.component.tsx
@@ -41,7 +41,7 @@ const btnSmallBase = `
 
 const btnXSBase = `
   jn:text-xs
-  jn:leading-normal
+  jn:leading-4
 `
 
 // default size typography
--- a/packages/ui-components/src/components/Button/Button.component.tsx
+++ b/packages/ui-components/src/components/Button/Button.component.tsx
@@ -39,7 +39,7 @@ const btnSmallBase = `
   jn:leading-5
 `
 
-const btnExtraSmallBase = `
+const btnXSBase = `
   jn:text-xs
 `
 
@@ -59,7 +59,7 @@ const btnSmallSubduedPadding = `
   jn:px-1.75
 `
 
-const btnExtraSmallDefaultPadding = `
+const btnXSDefaultPadding = `
   jn:py-1
   jn:px-1.5
 `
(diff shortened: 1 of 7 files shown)


TilmanHaupt (dismissed):

packages/ui-components/src/components/Button/Button.component.tsx
```diff
@@ -192,13 +229,15 @@ export const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonPr
     const titleValue = title || label || ""
 
     const buttonIcon = progress ? (
-      <Spinner size={size === "small" ? "1.125rem" : "1.5rem"} color={spinnerColorClass(variant)} />
+      <span className={`${label || children ? iconClasses(size) : ""}`}>
```
edda: This feels hacky in combination with the hard overwrite for the right margin on the Spinner (`jn:!mr-0`). Previously we didn't have custom spacing for the Spinner. If this becomes necessary for the xsmall button, instead of a forced overwrite make it possible to configure the inbuilt margin on the Spinner component via prop.

packages/ui-components/src/components/DataGridToolbar/DataGridToolbar.stories.tsx
```diff
@@ -53,25 +53,23 @@ export const ComplexCustomLayout: Story = {
...
+              <ComboBoxOption value="Electronics" />
+              <ComboBoxOption value="Clothing" />
+              <ComboBoxOption value="Furniture" />
+            </ComboBox>
+          </InputGroup>
           <Stack gap="2" wrap>
             <Pill pillKey="category" pillValue="electronics" closeable />
             <Pill pillKey="status" pillValue="active" closeable />
             <Pill pillKey="priority" pillValue="high" closeable />
+            <Button label="Remove filters" variant="subdued" size="extra-small" />
```
edda: Please change label to "Clear Filters" (always use title cases for buttons, clear instead of remove is shorter and feels less destructive)

packages/ui-components/src/components/Button/Button.component.tsx
```diff
@@ -55,6 +59,11 @@ const btnSmallSubduedPadding = `
   jn:px-1.75
 `
 
+const btnExtraSmallDefaultPadding = `
+  jn:py-1
+  jn:px-1
```
edda: The left/right padding feels a bit small compared to the top/bottom padding. Typically the x-padding has to be slightly larger than the y-padding because the line-height of the text adds additional spacing.

packages/ui-components/src/components/Button/Button.component.tsx
MartinS-git: All buttons had an issue — they became wider when the spinner appeared. This seemed incorrect, so I added a `marginRight` prop to the Spinner component to fix it.

packages/ui-components/src/components/Button/Button.component.tsx
```diff
@@ -39,6 +39,10 @@ const btnSmallBase = `
   jn:leading-5
 `
 
+const btnXSBase = `
+  jn:text-xs
```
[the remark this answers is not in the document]
edda: This is good feedback and makes sense, imo

edda (dismissed):
Other than the valid comment by Copilot regarding line-height for the xs button, looking very good now!

edda (approved):

TilmanHaupt (approved):

PR #1711, opened 2026-05-26
feat(ui): add `href`, `onClick` and `disabled` props to Card

### LINKED ISSUE #1623: [Task](ui): add `href` and `onClick`props to Card
Add `href` and `onClick` props to Card, and render interactive elements accordingly.

Allow Card to render as a native `<a>` or `<button>` element when `href` and/or `onClick` props are provided, following the established pattern used by `Button`, `NavigationItem`, and `SideNavigationItem`. Currently Card always renders as a `<div>` with no interactive semantics. Whne introducing this functionality, we should also add a `disabled` prop.

Clickable cards are a common UI pattern. Without this functionality, developers have to wrap or nest interactive elements inside a `<div>`, producing poor accessibility semantics and keyboard behavior.

## Props

Prop | Type | Default | Description
-- | -- | -- | --
`href` | `string` | `undefined` | When set, renders the card as an `<a>` element
`onClick` | `MouseEventHandler` | `undefined` | When set (without href), renders the card as a `<button>` element
`disabled` | `boolean` | `false` | Disables interaction; native disabled on `<button>`, aria-disabled + stripped href on `<a>`, renders "not-allowed"-cursor

## Heuristics What to Render

Props passed | Renders as
-- | --
neither `href` nor `onClick` | `<div>` (current behaviour, unchanged)
`onClick` only | `<button>`
`href` only | `<a>`
both `href` and `onClick` | `<a>` (href takes precedence)

## Implementation Notes

- Change base type from `HTMLAttributes<HTMLDivElement>` to `HTMLAttributes<HTMLElement>` to accommodate all three element types cleanly
- Add `href?: string` and `onClick` to the props interface explicitly
- Visual styles (`cardStyles`, `cardPadding`) are shared across all three variants — no change needed
- Disabled handling: on `<button>` use the native `disabled` attribute; on `<a>` remove the `href` attribute and add `aria-disabled="true"`, handle `event.preventDefault()` in the click handler, show `not-allowed`-cursor
- Add `forwardRef` — ref type will need casting (as `Ref<HTMLDivElement | HTMLAnchorElement | HTMLButtonElement>`) since the rendered element is dynamic
- `<button>` variant: add `type="button"` to prevent accidental form submission
- `<a>` variant: no role needed — it is already a link by semantics

## Acceptance Criteria

- [ ]  Without href or onClick, renders as `<div>` — no behaviour change for existing consumers
- [ ]  With onClick only, renders as `<button type="button">` with correct focus and keyboard behaviour
- [ ]  With href, renders as `<a href="...">` and navigates on click
- [ ]  With both, renders as `<a>` and fires `onClick` in addition to navigating
- [ ]  disabled correctly suppresses interaction on both `<button>` and `<a>` variants
- [ ]  Visual appearance is identical across all three variants
- [ ]  `forwardRef` works correctly for all three element types
- [ ]  Storybook stories cover all four rendering variants and the disabled state
- [ ]  Unit tests cover all four rendering variants and disabled behaviour
- [ ]  No breaking changes for existing consumers

guoda-puidokaite:
# Summary

Added `href`, `onClick` and `disabled` props to Card. Please see ACs.

# Related Issues

Closes https://github.com/cloudoperators/juno/issues/1623

# Testing Instructions

See Storybook preview.

# Checklist

# PR Manifesto


packages/ui-components/src/components/Card/Card.component.tsx
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
```
guoda-puidokaite, line 5: Note: Aligned all styling across variants, changing defaults.

- Removed blue link
- Made button left-aligned

Image

guoda-puidokaite (commented):
Note

TilmanHaupt (approved):

TilmanHaupt (changes requested):
Image

Somehow my links looks not like a link in my preview? 
TBD: I really like a color indication at hover about cards - sure hat we should not do this?

TilmanHaupt (dismissed):

packages/ui-components/src/components/Card/Card.component.tsx
TilmanHaupt: The screenshot confused me - its from before the changes. And without link color like in the preview is correct? :)

guoda-puidokaite: Exactly, this is before... showing the default behaviour we usually see e.g. blue link. I just mentioned it was all removed and aligned with default `Card` variant.

packages/ui-components/debug-storybook.log
```diff
@@ -0,0 +1,37 @@
+[09:11:17.503] [INFO] storybook v10.3.6
```
franzheidl: This file as a whole should not be comitted, it is a local dev artifact if I am not mistaken?

packages/ui-components/src/components/Card/Card.component.tsx
```diff
@@ -42,11 +58,58 @@ export interface CardProps extends HTMLAttributes<HTMLDivElement> {
...
+    }
+
+    if (href) {
+      // Render as <a> - takes precedence over button
+      return (
+        <a
+          ref={ref as Ref<HTMLAnchorElement>}
+          href={disabled ? undefined : href}
+          aria-disabled={disabled ? true : undefined}
+          className={`${combinedClassName} jn:block jn:w-full jn:overflow-hidden jn:whitespace-nowrap jn:text-ellipsis jn:truncate jn:text-inherit`}
```
franzheidl: The `<a>` has no native `disabled` attribute and will still be relevant in tab order even if `disabled` is set to `true`. You'd need explicitly ignore it in the taborder like so: `tabIndex={disabled ? -1 : undefined}` in order to exclude a disabled `<a>` from the tab order.

Also, `jn:truncate` may already include `no-wrap`, `overflow-hidden`, and `text-ellipsis`. You may want to double-check to avoid redundancy / code duplication.

packages/ui-components/src/components/Card/Card.component.tsx
```diff
@@ -42,11 +58,58 @@ export interface CardProps extends HTMLAttributes<HTMLDivElement> {
...
-export const Card = ({ children, padding = false, className = "", ...props }: CardProps): ReactNode => {
-  const combinedClassName = `juno-card ${cardStyles} ${padding ? cardPadding : ""} ${className}`
-  return (
-    <div className={combinedClassName} {...props}>
-      {children}
-    </div>
-  )
-}
+export const Card = forwardRef<HTMLElement, CardProps>(
+  ({ children, padding = false, className = "", href = "", onClick, disabled = false, ...props }, ref) => {
```
franzheidl: I think the correct default for `href` as an optional prop would be no default at all or `undefined`?

franzheidl (changes requested):

packages/ui-components/debug-storybook.log
guoda-puidokaite: Accidental, thanks. A new file since upgrade we need to `gitignore`.

packages/ui-components/src/components/Card/Card.component.tsx
guoda-puidokaite: Updated, thanks for the catches. Actually no need to truncate since we need to keep it all consistent.

packages/ui-components/src/components/Card/Card.component.tsx
```diff
@@ -42,11 +58,59 @@ export interface CardProps extends HTMLAttributes<HTMLDivElement> {
...
-  const combinedClassName = `juno-card ${cardStyles} ${padding ? cardPadding : ""} ${className}`
-  return (
-    <div className={combinedClassName} {...props}>
-      {children}
-    </div>
-  )
-}
+export const Card = forwardRef<HTMLElement, CardProps>(
+  ({ children, padding = false, className, href, onClick, disabled = false, ...props }, ref) => {
+    const combinedClassName = `juno-card ${cardStyles} ${padding ? cardPadding : ""} ${className} ${disabled ? "jn:cursor-not-allowed" : ""}`
```
franzheidl, line 63: (Regression caused by removing the empty string default from `className`, too)

If interpolated like this, removing `""` as a default from `className` would actually render `undefined` as a string in the `combinedClassName`, so `className` actually needs to use `""` as a default, or you'd need to guard the `className` right in the interpolation. My recommendation would be to use `""` as a default in this case, and not further complicate the interpolation.

franzheidl (changes requested):

franzheidl (approved):

TilmanHaupt (approved):

PR #1733, opened 2026-06-01
feat(ui): fix SideNavigation

guoda-puidokaite:
# Summary

Fixed issues in `SideNavigation`,`SideNavigationItem` and `SideNavigationGroup`:
- Removed `disabled` prop from `SideNavigation`
- Removed `activeItem` and `onActiveItemChange` prop from `SideNavigation`
- Removed `disabled` prop from `SidenavigationGroup`
- Fixed multiple issues in all components

# Changes Made

See ACs.

# Related Issues

Fixes https://github.com/cloudoperators/juno/issues/1573

# Checklist

# PR Manifesto


.changeset/clean-loops-look.md
```diff
@@ -0,0 +1,9 @@
+---
+"@cloudoperators/juno-ui-components": patch
```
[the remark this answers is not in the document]
guoda-puidokaite: Are we sure we want it as patch, @franzheidl?

TilmanHaupt (dismissed):

.changeset/clean-loops-look.md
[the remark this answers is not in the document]
edda: While the change is breaking, the prop has never worked. Though I agree, let's turn it into a major version and in the description use the keyword `Breaking:` and describe the breaking change. Even though the prop never worked properly if someone did use it they will get Typescript issues now.

packages/ui-components/src/components/SideNavigation/SideNavigation.component.tsx
```diff
@@ -24,39 +24,23 @@ const sideNavStyles = `
...
 export const SideNavigation = ({
-  activeItem = "",
   ariaLabel,
   children,
   className = "",
-  disabled = false,
-  onActiveItemChange,
   ...props
-}: SideNavigationProps): ReactNode => {
+}: Omit<SideNavigationProps, "activeItem" | "disabled" | "onActiveItemChange">): ReactNode => {
```
franzheidl: I think this `Omit` does not do anything since the props have been removed from the interface already anyway?

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
```diff
@@ -104,81 +95,81 @@ export const SideNavigationItem = ({
...
+    </div>
   )
 
   const combinedClassNames = `
     juno-sidenavigation-item
     ${sideNavItemStyles}
+    ${selected ? "juno-sidenavigation-item-selected" : ""}
+    ${isOpen ? "juno-sidenavigation-item-open" : ""}
+    ${disabled ? "jn:cursor-not-allowed" : "jn:cursor-pointer"}
     ${disabled ? disabledStyles : ""}
```
franzheidl, line 141: These two lines add yet another layer of declaring a disabled/not-allowed cursor, instead of removing one as the ticket states:

`disabledStyles` also contains `jn:cursor-not-allowed`, as does 
`.juno-sidenavigation-item:disabled { cursor: not-allowed; }` in the `sidenavigationitem.css` file.

Two of these should be removed.

packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
```diff
@@ -59,28 +48,27 @@ export const SideNavigationGroup = ({
...
-          size="24"
-          className={`
-            juno-sidenavigation-item
-            `}
-          icon={isOpen ? "expandMore" : "chevronRight"}
-        />
-      </span>
+      <button
+        onClick={handleToggleOpen}
+        className="expand-icon"
```
franzheidl: Is there any specific reason to add this class here? it does not appear to add any styling. 

If it is meant purely as an identifier it should at least be prefixed with `juno-` to match our pattern for identifier-classes, and ideally be more specific (such as `juno-sidenavigation-item-expand-icon`), although I'm not sure whether it makes a lot of sense to do that here.

Also, we should add `type="button"` here in order prevent defaulting to `type="submit"` under some circumstances.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
```diff
@@ -104,81 +95,81 @@ export const SideNavigationItem = ({
...
-          icon={isOpen ? "expandMore" : "chevronRight"}
-        />
-      </span>
+  const renderExpandButton = () =>
+    children && typeof children !== "string" && Children.count(children) ? (
+      <button
+        onClick={handleToggleOpen}
+        aria-label={isOpen ? "Collapse section" : "Expand section"}
+        disabled={disabled}
+        className="expand-icon"
```
franzheidl, line 122: See comment on `SideNavigationGroup`, same applies here.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
```diff
@@ -104,81 +95,81 @@ export const SideNavigationItem = ({
...
-          onClick={handleMainClick}
+          onClick={!disabled ? handleMainClick : undefined}
+          aria-disabled={disabled}
+          tabIndex={disabled ? -1 : undefined}
           {...props}
         >
           {renderItem()}
         </a>
       ) : (
         <button
```
franzheidl: We should add `type="button"` here in order prevent defaulting to `type="submit"` under some circumstances.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
```diff
@@ -104,81 +95,81 @@ export const SideNavigationItem = ({
...
       {href ? (
         <a
           aria-current={selected ? "page" : undefined}
-          aria-label={ariaLabel}
+          aria-label={ariaLabel ? ariaLabel : undefined}
           className={combinedClassNames}
           href={!disabled ? href : undefined}
-          onClick={handleMainClick}
+          onClick={!disabled ? handleMainClick : undefined}
+          aria-disabled={disabled}
```
franzheidl: This would render `aria-disabled="false"` on enabled `<a>` elements. This is not wrong, but adds unneccesary noise to the resulting markup, we should use `aria-disabled={disabled || undefined}` instead.

franzheidl (changes requested):

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
```diff
@@ -104,81 +95,81 @@ export const SideNavigationItem = ({
...
 
   const renderLeft = () => (
     <span className={leftStyles}>
       {icon && level === 0 ? <Icon className={"jn:mr-[0.25rem]"} icon={icon} size="24" /> : null}
-      {/* Add spacing before label which expands by level */}
       <div className={`level-${level + 1}`}>{label || children}</div>
     </span>
   )
 
   const renderItem = () => (
```
franzheidl: `renderItem()` composes the expand button into the subtree that is then placed wholesale inside the outer `<a>` / `<button>` (lines 174 and 184). This produces a `<button>` nested inside a `<button>` or inside an `<a>`, which is invalid HTML — browsers handle it inconsistently and focus/click behavior can become unreliable.

The expand button needs to be a sibling of the primary interactive element, not a descendant. The fix is to move `{renderExpandButton()}` out of `renderItem()` and render it alongside the outer `<a>` / `<button>` in a wrapper `<div>`, e.g. something like this:

```
<div className="jn:flex jn:w-full jn:items-center jn:justify-between">
  {href ? (
    <a ...>{renderLeft()}</a>
  ) : (
    <button ...>{renderLeft()}</button>
  )}
  {renderExpandButton()}
</div>
```

button-in-button

franzheidl (changes requested):

packages/ui-components/src/components/SideNavigation/SideNavigation.component.tsx
guoda-puidokaite: Thanks, cleaned up.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
guoda-puidokaite: I kept opacity in one of them, that's it. Thank You.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
guoda-puidokaite: Thank You for spotting.

TilmanHaupt (approved):

franzheidl (approved):
lgtm. Thank you, @guoda-puidokaite !

PR #1756, opened 2026-06-15
chore(ui): improve BreadcrumbItem

### LINKED ISSUE #1717: [Task](ui): refactor BreadcrumbItem to render semantically correct elements based on props
**Task Description**

`BreadcrumbItem` currently always renders an `<a>` element regardless of the props passed. This is semantically incorrect: if an item only has an `onClick` (no navigation destination), a `<button>` is the right element; if an item has neither `href` nor `onClick`, a plain `<span>` is appropriate. Additionally, `href` defaults to `"#"`, which means any non-active, non-disabled item without an explicit `href` produces a link that scrolls to the top of the page and pollutes browser history.

The `active` state also currently lacks `aria-current="page"`, which is the ARIA best practice for the current page item in a breadcrumb.

**Target rendering matrix:**

| Condition | Renders |
|---|---|
| `active` | `<span aria-current="page">` |
| `disabled` | `<span aria-disabled="true">` |
| `href` provided (not active/disabled) | `<a href={href} onClick={onClick}>` |
| `onClick` only, no `href` | `<button type="button" onClick={onClick}>` |
| neither `href` nor `onClick` | `<span>` |

`active` takes priority over `href` and `onClick` — the current page item is never interactive. When both `href` and `onClick` are passed, the component renders an `<a>` and attaches the click handler to it — links can have click handlers and this is the natural web behaviour. This matches how `Card` and `Icon` handle the same three-way rendering decision; `Card.component.tsx` can serve as the implementation reference.

**Sub-tasks**

- [x] Remove `href="#"` default — default to `undefined`
- [x] Add `<button type="button">` render path for `onClick`-only case
- [x] Use `<span>` as the "neither" fallback (consistent with `active`/`disabled` branches)
- [x] Remove the outer `<span>` wrapper around `<a>` — apply layout classes directly to the interactive element as root; when both `href` and `onClick` are passed, attach `onClick` to the `<a>`
- [x] Add `aria-current="page"` to the `active` branch
- [x] Fix disabled-on-`<a>` handling: strip `href`, add `aria-disabled={true}` and `tabIndex={-1}` (matches current/recent changes to `Card` component)
- [x] Widen `onClick` type from `MouseEventHandler<HTMLAnchorElement>` to `MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>`
- [x] Update stories to cover all render cases
- [x] Update tests accordingly

**Breaking changes to document:**
- Removing the `href="#"` default: items without an explicit `href` and without `onClick` now render as `<span>` instead of `<a href="#">`. This fixes a bug but is a behavioral change.
- Removing the outer `<span>` wrapper changes the DOM structure from `<span.juno-breadcrumb-item><a>…</a></span>` to `<a class="juno-breadcrumb-item">…</a>`. Any downstream CSS targeting the old structure (e.g. `.juno-breadcrumb-item > a`) would need updating.

guoda-puidokaite:
# Summary

Updated `BreadcrumbItem` with BREAKING CHANGES:
- Removed Default href="#": This eliminated default navigation behavior, changing <a> to <span> without explicit href.
- Structural Updates: Modified DOM structure by eliminating unnecessary <span> wrappers.
- Rendering Logic Enhanced: Introduced <button> for onClick without href. Ensured <span> usage for neither href nor onClick.
- Accessibility Improvements: Added aria-current="page" for active and aria-disabled for disabled items.
- Expanded onClick Type: Now supports both <a> and <button> elements for broader interaction handling.

# Changes Made

**See ACs** to verify changes made.

# Related Issues

Closes https://github.com/cloudoperators/juno/issues/1717

# Checklist

# PR Manifesto


packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.component.tsx
```diff
@@ -19,15 +19,15 @@ const breadcrumbLinkBaseStyles = `
...
+export interface BreadcrumbItemProps extends HTMLAttributes<HTMLElement> {
   /**
    * The type of icon to display within the breadcrumb item.
    */
   icon?: KnownIcons
 
   /**
    * The URL that the breadcrumb item points to for navigation.
-   * @default "#"
+   * @default undefined
```
andypf, line 25: nice 👍

andypf (dismissed):

edda (commented):
Blocker comment to give me time to review

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.component.tsx
```diff
@@ -99,30 +99,68 @@ export const BreadcrumbItem = ({
...
+    )
   }
 
-  // Render the breadcrumb item as a link if neither active nor disabled
-  return (
-    <span className={combinedClassName} {...props}>
-      <a href={href} className={breadcrumbLinkBaseStyles} aria-label={ariaLabel || label} onClick={handleLinkClick}>
+  if (href) {
+    // Render the breadcrumb item as a link if href is provided and apply breadcrumbLinkBaseStyles
+    const linkClassName = `${combinedClassName} ${breadcrumbLinkBaseStyles}`
```
edda: For the button type we render only `combinedClassName`. We should do the same here. Looking at the contents of `breadcrumbLinkBaseStyles` they are essentially duplicates of styles already present in `combineClassName`. In order to ensure the link and button styles don't drift apart let's remove the `breadcrumbLinkBaseStyles` constant and the application of them here and use only `combinedClassName`

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
```diff
@@ -29,7 +29,6 @@ describe("BreadcrumbItem", () => {
       render(<BreadcrumbItem icon="help" />)
       const imgElement = screen.getByRole("img")
       expect(imgElement).toBeInTheDocument()
-      expect(imgElement).toHaveAttribute("alt", "help")
     })
```
edda, line 33: Why did you remove this assertion? It's good to verify that the icon is rendered if we specifically use the `icon` prop for this test

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
```diff
@@ -76,46 +79,44 @@ describe("BreadcrumbItem", () => {
...
 
-    test("renders a disabled item as passed", async () => {
+    test("renders a disabled item with aria-disabled", async () => {
       const onClickSpy = vi.fn()
       render(<BreadcrumbItem href="#" disabled data-testid="breadcrumbitem" onClick={onClickSpy} />)
-      expect(screen.getByTestId("breadcrumbitem")).toBeInTheDocument()
-      expect(screen.getByTestId("breadcrumbitem")).toHaveClass("juno-breadcrumb-item-disabled")
+      const element = screen.getByTestId("breadcrumbitem")
+      expect(element).toBeInTheDocument()
+      expect(element).toHaveAttribute("aria-disabled", "true")
```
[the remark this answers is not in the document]
edda: I actually agree with this comment. We should assert that disabled items **don't** render as links to catch potential regressions, since the acceptance criteria states that in the disabled case we don't render a link at all.

edda (changes requested):

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.stories.tsx
```diff
@@ -84,14 +86,50 @@ export const Disabled = {
...
+  },
+}
+
+export const Link = {
+  render: Template,
+
+  parameters: {
+    docs: {
+      description: {
+        story: "A breadcrumb item rendered as a <a> with provided href",
```
edda: Html tags in the story are actually rendered as html. So the text of the story when you look at it in storybook actually reads: "A breadcrumb item rendered as a  using onClick for interaction." Replace the tag with html entities to make it render as intended.

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.stories.tsx
```diff
@@ -84,14 +86,50 @@ export const Disabled = {
...
+  },
+}
+
+export const Button = {
+  render: Template,
+
+  parameters: {
+    docs: {
+      description: {
+        story: "A breadcrumb item rendered as a <button> using onClick for interaction.",
```
edda: Same as link story above, replace tag with html entities to make it render correctly.

edda (changes requested):
Additonally, if you look at storybook the Breadcrumb page makes no mention of the different options for breadcrumb items, so if you only look at the main story and never at the `BreadcrumbItem` stories, you 'll miss this completely. As we did for other components that have children, let's add at least one story to the `Breadcrumb` page which contains a bunch of different types of breadcrumb items and in its description refers to the `BreadcrumbItems` stories.

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.component.tsx
franzheidl: Agree. Also, `cobinedClassName ` sets `jn:flex`, which is then overwritten with `jn:inline-flex` when `breadcrumbLinkBaseStyles` is applied, so merging these two seems to clarify things here as well.

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
```diff
@@ -76,46 +79,45 @@ describe("BreadcrumbItem", () => {
...
     })
 
-    test("renders a disabled item as passed", async () => {
+    test("renders a disabled item with aria-disabled", async () => {
       const onClickSpy = vi.fn()
       render(<BreadcrumbItem href="#" disabled data-testid="breadcrumbitem" onClick={onClickSpy} />)
-      expect(screen.getByTestId("breadcrumbitem")).toBeInTheDocument()
-      expect(screen.getByTestId("breadcrumbitem")).toHaveClass("juno-breadcrumb-item-disabled")
+      const element = screen.getByTestId("breadcrumbitem")
+      expect(element).toBeInTheDocument()
```
franzheidl, line 94: When assigning the element to a `const`as in the previous line, a matching element either does exist in the document, or the test will fail right away when attempting `const` assignment, so checking `expect(element).toBeInTheDocument()` is obsolete – line 92 can safely be removed.

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
```diff
@@ -62,10 +61,14 @@ describe("BreadcrumbItem", () => {
...
+      render(<BreadcrumbItem href="#" data-testid="breadcrumbitem" />)
+      expect(screen.getByRole("link")).toBeInTheDocument()
+    })
+
     test("does not render a link if 'href' is null", () => {
-      render(<BreadcrumbItem active data-testid="breadcrumbitem" />)
+      render(<BreadcrumbItem aria-label="breadcrumbitem" data-testid="breadcrumbitem" />)
       expect(screen.queryByRole("link")).not.toBeInTheDocument()
-      expect(screen.getByTestId("breadcrumbitem")).toBeInTheDocument()
     })
```
franzheidl, line 74: According to the logic, the `href` branch fires first in case both  `href` and `onClick` are passed (which is correct/as designed). We should explicitly test this behaviour by adding a new test (`"renders a link if both href and onClick are passed"`) in order to catch any regressions, and as it is not totally obvious.
Also the `onClick` handler should be tested for this case explicitly.

franzheidl (dismissed):

packages/ui-components/src/components/BreadcrumbItem/BreadcrumbItem.test.tsx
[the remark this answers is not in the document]
guoda-puidokaite: This was already resolved.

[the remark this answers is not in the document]
edda: Ok. But I'm still missing the assertion that we're not rendering an `<a>` if `disabled`

edda (approved):

taymoor89 (approved):

hodanoori (approved):

TilmanHaupt (approved):

PR #1794, opened 2026-06-24
fix(ui): fix NaN input in Pagination

### LINKED ISSUE #1793: [Bug](ui): when NaN input is entered into the Pagination field, NaN in printed
**Description**

When NaN input is entered into the Pagination field e.g "#", NaN in printed

Image

**Steps to Reproduce**

1. Enter a NaN value e.g. "#"
2. Notice NaN is printed in the input field

**Expected Behavior**

Default to the last available page.

**Environment**

- OS: [e.g., Windows, macOS, Linux]
- Browser: [e.g., Chrome, Firefox, Safari]
- Version: [e.g., 22]

guoda-puidokaite:
# Summary

When NaN input is entered, instead of printing `NaN`, the last available page is defaulted to.

# Related Issues

Closes https://github.com/cloudoperators/juno/issues/1793

# Checklist

# PR Manifesto

### DIFF
--- a/packages/ui-components/src/components/Pagination/Pagination.component.tsx
+++ b/packages/ui-components/src/components/Pagination/Pagination.component.tsx
@@ -184,7 +184,7 @@ export const Pagination = ({
       // Enforce minimum and maximum limits
       if (inputValue < 1) {
         inputValue = 1
-      } else if (controlTotalPage !== undefined && inputValue > controlTotalPage) {
+      } else if (isNaN(inputValue) || (controlTotalPage !== undefined && inputValue > controlTotalPage)) {
         inputValue = controlTotalPage
       }
     }
--- a/.changeset/eleven-buses-bet.md
+++ b/.changeset/eleven-buses-bet.md
@@ -1,0 +1,5 @@
+---
+"@cloudoperators/juno-ui-components": patch
+---
+
+Fix NaN input in Pagination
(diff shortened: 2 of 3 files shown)


packages/ui-components/src/components/Pagination/Pagination.component.tsx
```diff
@@ -184,7 +184,7 @@ export const Pagination = ({
       // Enforce minimum and maximum limits
       if (inputValue < 1) {
         inputValue = 1
-      } else if (controlTotalPage !== undefined && inputValue > controlTotalPage) {
+      } else if (isNaN(inputValue) || (controlTotalPage !== undefined && inputValue > controlTotalPage)) {
```
edda, line 187: hmm, this won't work if `controlTotalPage` is `undefined`. Imo a better way to deal with `NaN` values is to add it to the `inputValue < 1` case instead, because there is always going to be a first page.

ArtieReus (commented):
what about  adding following tests: 
  - What happens when totalPages is undefined and user enters non-numeric input
  - Edge cases like empty string, whitespace, special characters beyond just "$"

The test only covers the case where totalPages={5} is defined.

packages/ui-components/src/components/Pagination/Pagination.component.tsx
TilmanHaupt: !Number.isInteger(inputValue)  - is maybe also the wanted function to catch everything what is not a possible page. As 0.5 is not a possible page.

PR #1798, opened 2026-06-25
fix(ui): fix Select keyboard navigation — first arrow key press now highlights first item

### LINKED ISSUE #1781: [Task](ui): fix keyboard navigation in Select so first arrow key press focuses first menu item
Follow-up to #1748 / #1774.

When opening a `Select` via keyboard and navigating with arrow keys, several keystrokes are consumed before the first menu item appears focused. The menu items are there and selectable, but the focus ring does not appear until after a number of arrow presses.

## Summary

The Select menu's dropdown options were always present in the DOM (just hidden via `display: none`), which caused Headless UI to sort them by DOM position while they were invisible. Since hidden elements don't participate in layout, the sort produced unreliable results — so the first several arrow-down presses couldn't locate the correct option to highlight. The fix is to only render the dropdown into the DOM when the menu is actually open, so Headless UI always sorts against visible, laid-out elements.

---

## Root Cause

The original issue description (as per below) pointed at `tabIndex` on the Floating UI container as the culprit — that diagnosis turned out to be misleading. The real cause is deeper in how Headless UI's internal state machine interacts with the portal.

### What was actually happening

The `ListboxOptions` (and all `SelectOption` children) were always mounted in the DOM, hidden only via a `display: none` style on the wrapping Floating UI container div. This means on page load, before the menu is ever opened, HUI's option registration cycle runs:

1. Each `ListboxOption` registers itself with the machine via a batched microtask (`RegisterOptions`)
2. The machine schedules a `SortOptions` action via `requestAnimationFrame` to sort the registered options by their DOM position (`sortByDomNode` / `compareDocumentPosition`)
3. The sort runs against elements that are inside a `display: none` subtree — `compareDocumentPosition` can return 0 for all comparisons in this state, making the sort a no-op or producing an undefined order

When the menu is then opened and the user presses arrow-down, `goToOption({ focus: Next })` fires. Because `activeOptionIndex` is null, HUI calls `g(state)` which re-runs `sortByDomNode` — again against elements whose DOM context was established while hidden. `calculateActiveIndex` gets a list it can't reliably order, so the resulting index is wrong or null. The first few arrow-down presses keep returning null or an incorrect index, with no `data-active` applied — until enough state has settled that the sort resolves correctly.

This explains why pressing Enter after a single arrow-down selected the correct item (HUI's logical state eventually resolved) but the visual `data-active` attribute didn't appear until press 4: the machine state was updating correctly but `isActive` was returning false for the wrong option index.

## The Fix

Conditionally render the portal only when the menu is open (`{open && createPortal(...)}`) and remove the `display` toggling style. This means options only mount, register, and sort when the menu is actually open and the container is visible in the DOM — so `compareDocumentPosition` has a real layout to work with, and `goToOption` always finds a correctly-sorted option list.

The `display: none` / always-mounted approach was likely chosen to avoid the portal unmounting/remounting on every open/close, but it inadvertently broke HUI's DOM-order-dependent focus management.

## Side note on the `tabIndex` change

The `tabIndex={-1}` on the floating container is still correct and worth keeping. Without it, `getFloatingProps()` leaves the container participating in sequential focus order, which can cause its own navigation issues — it just wasn't the cause of the 4-press lag.

### _Root Cause_
_The `Select` menu is rendered inside a portal using Floating UI. The floating container `<div>` receives `getFloatingProps()` from `@floating-ui/react`'s `useInteractions`, which spreads interaction props — including a `tabIndex` — onto the div. This places the container and/or the `ListboxOptions` element into the browser's sequential focus order. Arrow key presses are consumed navigating through these invisible focusable elements before Headless UI's virtual list navigation kicks in for the items._

_### Fix_
_Add `tabIndex={-1}` explicitly to the floating container `<div>` in `Select.component.tsx` so it can still receive programmatic focus (required by Floating UI) but is excluded from sequential keyboard navigation._

## Acceptance criteria

- [ ] Opening a Select via keyboard and pressing arrow down immediately highlights the first item
- [ ] Opening a Select via mouse and pressing arrow down immediately highlights the first item
- [ ] Existing keyboard open/close/select behaviour does not regress

franzheidl:
**DO NOT MERGE YET – SEE COMMENT**

## Problem

When using the keyboard to navigate the menu of a `Select`, it takes hitting the `arrow-down`-key multiple times before the first item was visually focussed. With the hits before items were focussed, too, but without any visual display, which is highly confusing.

### Root Cause

Dropdown options were always mounted in the DOM, hidden via `display: none` on the Floating UI container. Headless UI's DOM-order sort ran against elements in a hidden subtree on page load, producing an unreliable option order. On first open, `goToOption` couldn't resolve the correct active index and the first few arrow-down presses had no visible effect.

## Fix

- Conditionally render the portal only when open (`{open && createPortal(...)}`)
  so options mount, register, and sort against visible DOM elements
- Switch `SelectOption` from `as={Fragment}` to `as="li"` so HUI owns the
  element and sets `data-active` natively
- Add `tabIndex={-1}` to the floating container to exclude it from sequential
  keyboard focus order

Closes #1781

## Test plan
- [ ] Open Select via keyboard, first arrow-down immediately highlights item 1
- [ ] Open Select via mouse, item 1 (or selected item) is highlighted immediately on open
- [ ] Arrow-up/down, Enter, Escape, Tab all behave correctly
- [ ] Multi-select keyboard navigation works
- [ ] No visual regressions in Select stories

# Checklist

# PR Manifesto

### DIFF
--- a/packages/ui-components/src/components/Select/Select.component.tsx
+++ b/packages/ui-components/src/components/Select/Select.component.tsx
@@ -418,33 +418,34 @@ export const Select = ({
                       )
                     }}
                   </ListboxButton>
-                  {createPortal(
-                    <div
-                      ref={refs.setFloating}
-                      className={`
+                  {open &&
+                    createPortal(
+                      <div
+                        ref={refs.setFloating}
+                        className={`
                           juno-select-menu-container
                           ${menuStylesContainer}
                         `}
-                      style={{
-                        position: strategy,
-                        top: y ?? 0,
-                        left: x ?? 0,
-                        display: open ? "block" : "none",
-                      }}
-                      {...getFloatingProps()}
-                    >
-                      <ListboxOptions
-                        static
-                        className={`
-                          juno-select-menu
-                          ${menuStyles}
-                        `}
+                        style={{
+                          position: strategy,
+                          top: y ?? 0,
+                          left: x ?? 0,
+                        }}
+                        {...getFloatingProps()}
+                        tabIndex={-1}
                       >
-                        {children}
-                      </ListboxOptions>
-                    </div>,
-                    portalContainerRef ?? document.body
-                  )}
+                        <ListboxOptions
+                          static
+                          className={`
+                            juno-select-menu
+                            ${menuStyles}
+                          `}
+                        >
+                          {children}
+                        </ListboxOptions>
+                      </div>,
+                      portalContainerRef ?? document.body
+                    )}
                 </div>
               </>
             )
--- a/packages/ui-components/src/components/SelectOption/SelectOption.component.tsx
+++ b/packages/ui-components/src/components/SelectOption/SelectOption.component.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { Fragment, HTMLAttributes, ReactNode, useContext, useEffect } from "react"
+import React, { HTMLAttributes, ReactNode, useContext, useEffect } from "react"
 import { ListboxOption } from "@headlessui/react"
 import { SelectContext } from "../Select/Select.component"
 import { Icon } from "../Icon"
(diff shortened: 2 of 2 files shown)


franzheidl:
**DO NOT MERGE – RAW VALUES IN LABELS REGRESSION:**

**Potential regression: pre-selected label shows raw `value` until dropdown is first opened**

With `{open && createPortal(...)}`, `SelectOption` children only mount when the dropdown opens. Their `useEffect` that calls `addOptionValueAndLabel` therefore doesn't fire until first open, leaving `optionValuesAndLabels` empty on initial render.

`getDisplayValue` (line 465) falls back to `valueLabel || value || placeholder` when the map has no entry for the current value — so a controlled `<Select value="opt-1">` with `<SelectOption value="opt-1" label="Option One" />` will render `"opt-1"` in the toggle button on load instead of `"Option One"`. Multi-select is worse: `getMultipleDisplayValues` returns an empty string for all pre-selected options before first open.

One clean fix would be to replace the effect-based registration with a `useMemo` that walks `children` directly using `React.Children.forEach` at render time, so the map is populated independently of whether the portal has mounted. This would also eliminate the N `setState` calls per open cycle as a side effect, since `addOptionValueAndLabel` and its `useEffect` in `SelectOption` could be removed entirely.

---

**Minor: `tabIndex={-1}` placed after `{...getFloatingProps()}` silently overrides any `tabIndex` Floating UI emits**

As of the current Floating UI setup (`useClick` + `useDismiss`) neither hook emits a `tabIndex`, so this is benign right now. But with `tabIndex={-1}` appearing after the spread, any future interaction hook (e.g. `useListNavigation`) that sets a `tabIndex` for focus tracking would be silently clobbered. Placing `tabIndex={-1}` before the spread, or passing it through the Floating UI API as `{...getFloatingProps({ tabIndex: -1 })}`, is the safer pattern.

PR #1803, opened 2026-06-29
feat(ui): add onToggle callback to SideNavigationGroup and SideNavigationItem

### LINKED ISSUE #1799: [Feature](ui): Make SideNavigationGroup a Controlled Component
## Problem

The `SideNavigationGroup` component from `@cloudoperators/juno-ui-components` is currently **uncontrolled**, meaning it manages its own internal open/closed state without exposing any way for parent components to control or observe that state.

This prevents implementing common UX patterns such as:
- Automatically opening a section when navigating to a service within it
- Programmatically controlling which sections are open/closed
- Persisting section state across navigation
- Responding to section toggle events

## Current Behavior

```tsx
// Current usage - props are ignored
<SideNavigationGroup 
  label="Compute"
  open={true}  
  onToggle={(isOpen) => console.log(isOpen)}  // ❌ This callback is never called
>
  {/* children */}
</SideNavigationGroup>
```

The component:
- Does not call any `onToggle` callback
- Maintains internal state that cannot be accessed from outside

## Requested Changes

Make `SideNavigationGroup` a **controlled component** by:

### 1. Respect the `open` prop

When `open` prop is provided, the component should use it to determine its expanded/collapsed state:

```tsx
<SideNavigationGroup 
  label="Compute"
  open={isComputeOpen}  
>
  {/* children */}
</SideNavigationGroup>
```

### 2. Add `onToggle` callback

When the user clicks to expand/collapse the section, call the `onToggle` callback:

```tsx
<SideNavigationGroup 
  label="Compute"
  open={isComputeOpen}
  onToggle={(isOpen: boolean) => {
    setIsComputeOpen(isOpen)
  }}
>
  {/* children */}
</SideNavigationGroup>
```

### 3. Support both controlled and uncontrolled modes

For backward compatibility, the component should support both modes:

**Controlled mode** (when `open` prop is provided):
```tsx
<SideNavigationGroup 
  label="Compute"
  open={isOpen}
  onToggle={setIsOpen}
>
```

**Uncontrolled mode** (when `open` prop is omitted):
```tsx
<SideNavigationGroup 
  label="Compute"
  defaultOpen={true}  // Optional: control initial state
>
```

## Proposed TypeScript Interface

```typescript
interface SideNavigationGroupProps {
  label: string
  children: React.ReactNode
  
  // Controlled mode
  open?: boolean
  onToggle?: (isOpen: boolean) => void
  
  // Uncontrolled mode
  defaultOpen?: boolean
}
```

## Use Case

We need to implement the following behavior in Aurora Dashboard:

1. **Default**: All sections are open
2. **User closes a section**: It stays closed during navigation
3. **Navigate to a service in a closed section**: Automatically open that section
4. **Other sections**: Never change unless user manually toggles them

Without a controlled `SideNavigationGroup`, we cannot implement this UX pattern.

## Alternative Workaround

If controlled behavior cannot be added, at minimum provide:
- A `ref` API to programmatically control the component
- An event listener or callback for state changes

## Priority

**Medium** - This blocks implementing important navigation UX improvements in Aurora Dashboard.

## Related Patterns

This follows the standard React pattern for form controls:
- Controlled: `<input value={x} onChange={setX} />`
- Uncontrolled: `<input defaultValue={x} />`

References:
- [React: Controlled Components](https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable)
- [React: Uncontrolled Components](https://react.dev/reference/react-dom/components/input#providing-an-initial-value-for-an-input)

edda:
## Summary

- Adds `onToggle?: (isOpen: boolean) => void` to `SideNavigationGroup` and `SideNavigationItem`. The callback fires whenever the user clicks the toggle affordance, with the next open state.
- For `SideNavigationGroup` the whole row is the toggle, so `onToggle` fires on any click of the group. For `SideNavigationItem` only the chevron is the toggle — the label keeps its existing `onClick`/`href` navigation behavior unchanged.
- The existing `open` prop semantics are preserved: it still seeds the initial state and re-syncs the internal state when the parent re-renders with a new value. No breaking changes — consumers that don't pass `onToggle` see identical behavior to today.

This is the same soft-control pattern already used by `Panel` and `Modal` in this library and lets parents implement use cases like "auto-open the group containing the active route while preserving any groups the user has manually closed" (the Aurora Dashboard use case from the issue).

Closes #1799.

## Test plan

- [x] `pnpm vitest run` in `packages/ui-components` for `SideNavigation*` suites — 47/47 passing, including 13 new tests:
  - Group: `onToggle` fires with `true`/`false`, internal state still toggles without callback, `open` prop re-sync still works.
  - Item: `onToggle` fires from the chevron with the next state; label clicks do **not** fire `onToggle` (only `onClick`); disabled items don't fire `onToggle`; internal state still toggles without callback; `open` prop re-sync still works.
- [x] `pnpm --filter @cloudoperators/juno-ui-components typecheck` clean.
- [x] `pnpm --filter @cloudoperators/juno-ui-components lint` clean.
- [x] Storybook stories added: `Controlled` example for both components demonstrating the parent-owned-state pattern. Verify visually in Storybook.
- [ ] Reviewer to verify the new `Controlled` stories in Storybook behave as expected.

## Notes

- The component-level interface `SideNavigationItemProps` now uses `Omit<HTMLAttributes<HTMLElement>, "onToggle">` to avoid colliding with the DOM `ToggleEvent` handler type on `HTMLAttributes`. Our `onToggle(isOpen: boolean)` is a domain-level callback, not a DOM event handler.
- Includes a changeset entry (`minor`).

### DIFF
--- a/packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
+++ b/packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
@@ -51,8 +51,10 @@ export interface SideNavigationGroupProps {
   children?: ReactElement<SideNavigationItemProps> | ReactElement<SideNavigationItemProps>[]
   /** Label displayed for the navigation group. */
   label: ReactNode
-  /** Controls whether the navigation group is expanded by default. */
+  /** Controls whether the navigation group is expanded. When changed by the parent the group re-syncs to the new value. */
   open?: boolean
+  /** Fired when the user clicks the group to toggle it. Receives the next open state. */
+  onToggle?: (_isOpen: boolean) => void
 }
 
 /**
@@ -66,7 +68,12 @@ export interface SideNavigationGroupProps {
  * @see {@link SideNavigationGroupProps}
  **/
 
-export const SideNavigationGroup = ({ children, label = "", open = false }: SideNavigationGroupProps): ReactNode => {
+export const SideNavigationGroup = ({
+  children,
+  label = "",
+  open = false,
+  onToggle,
+}: SideNavigationGroupProps): ReactNode => {
   const [isOpen, setIsOpen] = useState(open)
 
   // Sync internal state with external prop changes
@@ -76,7 +83,9 @@ export const SideNavigationGroup = ({ children, label = "", open = false }: Side
 
   const handleToggleOpen = (e: MouseEvent<HTMLElement>) => {
     e.stopPropagation() //Prevent event bubbling when expanding/collapsing
-    setIsOpen(!isOpen)
+    const next = !isOpen
+    setIsOpen(next)
+    onToggle?.(next)
   }
 
   const hasChildren = !!children && Children.count(children) > 0
--- a/packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
+++ b/packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
@@ -51,7 +51,7 @@ export interface SideNavigationGroupProps {
   children?: ReactElement<SideNavigationItemProps> | ReactElement<SideNavigationItemProps>[]
   /** Label displayed for the navigation group. */
   label: ReactNode
-  /** Controls whether the navigation group is expanded. When changed by the parent the group re-syncs to the new value. */
+  /** Sets the open state of the navigation group. The component owns the open state internally but re-syncs to this prop whenever the parent updates it, so it can be used either as the initial value or to drive the state from the outside. */
   open?: boolean
   /** Fired when the user clicks the group to toggle it. Receives the next open state. */
   onToggle?: (_isOpen: boolean) => void
--- a/packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
+++ b/packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
@@ -39,7 +39,7 @@ const disabledStyles = `
   jn:opacity-50
 `
 
-export interface SideNavigationItemProps extends HTMLAttributes<HTMLElement> {
+export interface SideNavigationItemProps extends Omit<HTMLAttributes<HTMLElement>, "onToggle"> {
   /** Provides an accessibility label for the navigation item. */
   ariaLabel?: string
   /** Nested SideNavigationItem components rendered as a sub-list when expanded. A string may be passed instead and will be treated as a label. */
@@ -54,7 +54,9 @@ export interface SideNavigationItemProps extends HTMLAttributes<HTMLElement> {
   label?: ReactNode
   /** Function handler triggered upon item click. */
   onClick?: MouseEventHandler<HTMLElement>
-  /** Controls whether the item is expanded by default. */
+  /** Fired when the user clicks the chevron to expand or collapse the nested children. Receives the next open state. */
+  onToggle?: (_isOpen: boolean) => void
+  /** Controls whether the item is expanded. When changed by the parent the item re-syncs to the new value. */
   open?: boolean
   /** Indicates if the item is currently selected or active. */
   selected?: boolean
@@ -86,6 +88,7 @@ export const SideNavigationItem = ({
   icon,
   label = "",
   onClick,
+  onToggle,
   open = false,
   selected = false,
   ...props
@@ -100,7 +103,9 @@ export const SideNavigationItem = ({
   const handleToggleOpen = (e: MouseEvent<HTMLElement>) => {
     if (disabled) return
     e.stopPropagation()
-    setIsOpen(!isOpen)
+    const next = !isOpen
+    setIsOpen(next)
+    onToggle?.(next)
   }
 
   const handleMainClick = (e: MouseEvent<HTMLElement>) => {
--- a/packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
+++ b/packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
@@ -56,7 +56,7 @@ export interface SideNavigationItemProps extends Omit<HTMLAttributes<HTMLElement
   onClick?: MouseEventHandler<HTMLElement>
   /** Fired when the user clicks the chevron to expand or collapse the nested children. Receives the next open state. */
   onToggle?: (_isOpen: boolean) => void
-  /** Controls whether the item is expanded. When changed by the parent the item re-syncs to the new value. */
+  /** Sets the open state of the nested children. The component owns the open state internally but re-syncs to this prop whenever the parent updates it, so it can be used either as the initial value or to drive the state from the outside. */
   open?: boolean
   /** Indicates if the item is currently selected or active. */
   selected?: boolean
--- a/packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.test.tsx
+++ b/packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.test.tsx
@@ -173,9 +173,7 @@ describe("SideNavigationItem", () => {
       </SideNavigationItem>
     )
 
-    // The label button is the first button; the chevron is the second.
-    const buttons = screen.getAllByRole("button")
-    fireEvent.click(buttons[0])
+    fireEvent.click(screen.getByRole("button", { name: "Messages" }))
     expect(onClick).toHaveBeenCalledTimes(1)
     expect(onToggle).not.toHaveBeenCalled()
   })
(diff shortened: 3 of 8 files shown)


packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
```diff
@@ -54,7 +54,9 @@ export interface SideNavigationItemProps extends HTMLAttributes<HTMLElement> {
   label?: ReactNode
   /** Function handler triggered upon item click. */
   onClick?: MouseEventHandler<HTMLElement>
-  /** Controls whether the item is expanded by default. */
+  /** Fired when the user clicks the chevron to expand or collapse the nested children. Receives the next open state. */
+  onToggle?: (_isOpen: boolean) => void
+  /** Controls whether the item is expanded. When changed by the parent the item re-syncs to the new value. */
   open?: boolean
```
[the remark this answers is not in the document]
edda: Keeping the leading underscore: it matches the established convention in this codebase, enforced by our ESLint `no-unused-vars` rule (`varsIgnorePattern: /^_/`). Existing public callbacks here follow the same pattern, e.g. `ComboBox` (`onChange?: (_value: string) => void`) and `useTooltip` (`onOpenChange?: (_open: boolean) => void`). Changing it just for this prop would create an inconsistency. Updated the `open` JSDoc to make the seed+resync semantics explicit in 16ebca8.

packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
```diff
@@ -51,8 +51,10 @@ export interface SideNavigationGroupProps {
   children?: ReactElement<SideNavigationItemProps> | ReactElement<SideNavigationItemProps>[]
   /** Label displayed for the navigation group. */
   label: ReactNode
-  /** Controls whether the navigation group is expanded by default. */
+  /** Controls whether the navigation group is expanded. When changed by the parent the group re-syncs to the new value. */
   open?: boolean
+  /** Fired when the user clicks the group to toggle it. Receives the next open state. */
+  onToggle?: (_isOpen: boolean) => void
```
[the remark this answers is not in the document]
edda: Same response as on the Item: the leading underscore is the codebase convention enforced by ESLint's `varsIgnorePattern: /^_/` (matches `ComboBox.onChange`, `useTooltip.onOpenChange`, etc.), so keeping it for consistency. The `open` JSDoc is reworded in 16ebca8 to make the seed+resync semantics explicit.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.test.tsx
```diff
@@ -146,4 +146,75 @@ describe("SideNavigationItem", () => {
...
+      <SideNavigationItem label="Messages" onClick={onClick} onToggle={onToggle}>
+        <SideNavigationItem label="Inbox" />
+      </SideNavigationItem>
+    )
+
+    // The label button is the first button; the chevron is the second.
+    const buttons = screen.getAllByRole("button")
+    fireEvent.click(buttons[0])
+    expect(onClick).toHaveBeenCalledTimes(1)
+    expect(onToggle).not.toHaveBeenCalled()
```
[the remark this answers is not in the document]
edda: Good catch — fixed in 16ebca8. The test now queries the label button by its accessible name (`getByRole("button", { name: "Messages" })`) instead of by array index.

packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
```diff
@@ -76,7 +83,9 @@ export const SideNavigationGroup = ({ children, label = "", open = false }: Side
 
   const handleToggleOpen = (e: MouseEvent<HTMLElement>) => {
     e.stopPropagation() //Prevent event bubbling when expanding/collapsing
-    setIsOpen(!isOpen)
+    const next = !isOpen
```
MartinS-git, line 86: My AI flagged a potential issue with the handleToggleOpen implementation in both components.

The current code reads isOpen from the render-time closure to compute next:

```
const next = !isOpen
setIsOpen(next)
onToggle?.(next)
```
In React 18 with automatic batching or rapid clicks, two toggle events can fire before React commits a re-render. Both closures capture the same isOpen value, so both compute the same next — the second click is effectively a no-op, and onToggle reports the wrong value to the parent twice.

For the Aurora Dashboard use case ("auto-open the group containing the active route"), this means the parent could receive true, true instead of true, false and end up with incorrect open state.

The fix would be to compute next inside a functional updater, so it always reads the latest state:


```
setIsOpen(prev => {
  const next = !prev
  onToggle?.(next)
  return next
})
```
This guarantees onToggle always gets the correct next value, even under batching.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
```diff
@@ -100,7 +103,9 @@ export const SideNavigationItem = ({
   const handleToggleOpen = (e: MouseEvent<HTMLElement>) => {
     if (disabled) return
     e.stopPropagation()
-    setIsOpen(!isOpen)
+    const next = !isOpen
```
MartinS-git, line 106: Same as in SideNavigationGroup.component.tsx:
My AI flagged a potential issue with the handleToggleOpen implementation in both components.

The current code reads isOpen from the render-time closure to compute next:

```
const next = !isOpen
setIsOpen(next)
onToggle?.(next)
```
In React 18 with automatic batching or rapid clicks, two toggle events can fire before React commits a re-render. Both closures capture the same isOpen value, so both compute the same next — the second click is effectively a no-op, and onToggle reports the wrong value to the parent twice.

For the Aurora Dashboard use case ("auto-open the group containing the active route"), this means the parent could receive true, true instead of true, false and end up with incorrect open state.

The fix would be to compute next inside a functional updater, so it always reads the latest state:


```
setIsOpen(prev => {
  const next = !prev
  onToggle?.(next)
  return next
})
```
This guarantees onToggle always gets the correct next value, even under batching.

packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
TilmanHaupt: very nice.

franzheidl (approved):
LGTM!

Two (very) minor issues that we can easily address later (if at all):

- `_isOpen` underscore prefix is unnecessary in a public interface, just `isOpen` would be cleaner IMO?
-  mixed usage of `fn()` vs `action()` in some stories may raise some question marks,  and feels somewhat inconsistent

Neither of these two are anywhere near blocking though.

packages/ui-components/src/components/SideNavigationGroup/SideNavigationGroup.component.tsx
edda: Technically true, but probably not a problem in the real world since people would have to click very fast for the clicks to land within the same event tick. We're updating state in a similar manner in other components. So if we do tackle this, I would do a more global refactor out of it. Lastly I believe the proposed fix violates React's rule that state updaters must be pure (i.e. no side effects). Calling `onToggle?.(next)` within the update function violates this rule.

packages/ui-components/src/components/SideNavigationItem/SideNavigationItem.component.tsx
edda: Technically true, but probably not a problem in the real world since people would have to click very fast for the clicks to land within the same event tick. We're updating state in a similar manner in other components. So if we do tackle this, I would do a more global refactor out of it. Lastly I believe the proposed fix violates React's rule that state updaters must be pure (i.e. no side effects). Calling onToggle?.(next) within the update function violates this rule.

edda:
> LGTM!
> 
> Two (very) minor issues that we can easily address later (if at all):
> 
> * `_isOpen` underscore prefix is unnecessary in a public interface, just `isOpen` would be cleaner IMO?
> * mixed usage of `fn()` vs `action()` in some stories may raise some question marks,  and feels somewhat inconsistent
> 
> Neither of these two are anywhere near blocking though.

`isOpen` without underscore triggered the `no-unused-vars` linting rule. With underscore the variable is ignored for this rule. We did it in that way in ComboBox and Tooltip already, that's why I used the same here.

Mixed usage of `fn()` and `action()` was because the `fn()` spy didn't work on nested children so you couldn't see the action being triggered in storybook. But agree, if we spend a bit more time on it we can probably find a better way to do this that is less confusing.

MartinS-git (approved):

PR #1859, opened 2026-08-03
feat(ui): improve code-block-footer api

### LINKED ISSUE #1844: [Task](ui): improve `CodeBlockFooter´ API and copy confirmation
## Summary

This issue deals with two separate, but closely related topics:

## Copy confirmation is lost when customizing the footer content

`CodeBlockFooter` currently does not accept `children`, so consumers using the `codeBlockFooter` slot end up re-implementing the Copy button, confirmation etc. themselves. The `CustomCodeBlockFooter` story illustrates this.

Add `children?: ReactNode` to `CodeBlockFooter`. Children render to the left of the Copy button, which is always present and rightmost. 
Consumers who want custom footer content pass a `<CodeBlockFooter>` with children into the `codeBlockFooter` slot rather than building their own footer from scratch. The `copy` prop is preserved on the footer and respected, so the Copy button will be rendered as the rightmost button in the footer or not, depending on the value of the `copy` prop even when customizing the footer.

Update the `CustomCodeBlockFooter` story accordingly.

## Inline "Copied!" confirmation disrupts layout

When preserving the Copy button as described above along with the current in-place confirmation, theer is a conflict, as custom options and the confirmation will need to go to the same location:

The `<span>` showing `"Copied!"` shifts footer layout on each copy action and will push custom children around. Replace it with a `Tooltip` on the Copy button, driven by `isCopied`, visible for the same 1000 ms. The tooltip floats above the layout and thus does not affect footer dimensions nor interferes with any other (custom) options.

## Sub-tasks

- [ ] Add `children?: ReactNode` prop to `CodeBlockFooter`; render children left, Copy button (if poresent as per `Copy` prop) right
- [ ] Replace inline `"Copied!"` span with a `Tooltip` "Copied!" on the Copy button
- [ ] Update the `CustomCodeBlockFooter` story to pass a `<CodeBlockFooter>` with a Download icon as children, no Copy button
- [ ] Add a `CustomCodeBlockFooterWithCopy` story that sets `Copy` to true to illustrate how custom content (can be "Download", too) and Copy button can coexist
- [ ] Update tests to reflect children rendering order and tooltip-based confirmation

**Related Issues**

This issue is a follow-up to #1829

vlad-schur-external-sap:
# Summary

1. Copy confirmation is applied when customizing the footer content.
2. Inline "Copied!" confirmation does not disrupts the   layout.

# Changes Made

- Refactors CodeBlock to always render a unified CodeBlockFooter, passing codeBlockFooter as children.
- Updated CodeBlockFooter to use tooltip for copied action.
- Adjusts unit tests and Storybook stories to reflect the new footer composition behavior.

# Related Issues

- Issue 1844: cloudoperators/juno/issues/1844

# Testing Instructions

1. `pnpm i`
3. `pnpm TASK`

# Checklist

# PR Manifesto


franzheidl:
Overall LGTM!

Some notes I am not sure about:

- Should we export the `CodeBlockFooter` from the index file, too?
- Strictly speaking, the reworked footer is a breaking change – instead of passing a whole new footer, we inject content into the footer now. Not sure if this warrants a `major` in the changeset, would be good to document/note either way.
- In the `CodeBlock` test, we now import the component directly form the component file, which makes it consistent with the rest / most of our tests. Using this strategy however wouldn't catch an error in case somebody removed the export from an `index` file. This is more of a general topic rather than for this specific component or PR though, that we should discuss at some point.

franzheidl (dismissed):
LGTM, see my my other comment.

franzheidl (approved):
LGTM

hodanoori (approved):

edda (approved):

